PR #1, opened 2021-03-10
DDD-1: Support for multiple connector tasks for relational Debezium connectors

morozov:


DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+Currently, it is challenging to use Debezium Connector for SQL Server in a multi-tenanted environment where a single SQL Server instance hosts tens to hundreds of similar tenant-scoped databases due to the fact that a given connector instance can capture changes only from a single database. The challenges are:
+
+1. Kafka Connect resources. Deploying hundreds of connector instances will require a Kafka Connect cluster with significant CPU and memory resources.
+2. Cluster stability. If the resources above are insufficient (e.g. memory), it may cause a Connect worker crash which in turn will fail all connector tasks scheduled to the worker. In the best case scenario, those tasks can be just restarted. In a worse case scenario, the restarted connectors will produce duplicates since the source connectors currently don't implement exactly-once processing semantics. In an even worse case, a worker can fail during a long-running snapshot, handling which will require additional cleanup.
+3. SQL Server resources. Each server instance will have to maintain an extra set of persistent connections (one per database, tens to hundreds in total) which may require additional server resources.
```
gunnarmorling, line 9: So this one wouldn't be mitigated by support for multiple tasks per-se, right? Rather via the ability to capture changes from multiple databases on one SQL Server host using one task.

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Proposed Changes
+
+### Source partition awareness at the task level ([#15](https://github.com/sugarcrm/debezium/pull/15))
+
+Currently, a Debezium connector task can process a single source partition throught its lifecycle. This partition is usually hard-coded in the context initialization code. E.g.:
+
+```java
+public SqlServerOffsetContext(SqlServerConnectorConfig connectorConfig) {
+        partition = Collections.singletonMap(SERVER_PARTITION_KEY, connectorConfig.getLogicalName());
```
gunnarmorling, line 19: This will require some consideration when migrating existing connectors from the old offset format to the new one.

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+In order to achieve partition-awareness, we need to introduce an API and a set of components that do the following:
+
+1. Describe a *set* of source partitions to be processed by the task.
```
gunnarmorling, line 24: Would a source partition be a single database? Or table?

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Support for multiple tasks per connector ([#9](https://github.com/sugarcrm/debezium/issues/9))
+
+Apart from being able to process multiple databases by a given task, it would be great if a connector was able to start multiple tasks and divide the databases between them. This way, it would allow to scale a single connector just by changing its `max.tasks` configuration.
+
+Being able to specify which subset of the databases a given task should process warrants the introduction of a new task-level configuration parameter ([#17](https://github.com/sugarcrm/debezium/issues/17)):
+
+| Name                  | Type     | Default | Importance | Doc                                                          |
+| :-------------------- | :------- | :------ | :--------- | :----------------------------------------------------------- |
+| `task.database.names` | `STRING` | `None`  | `HIGH`     | The list of comma-separated database names form which the given connector task should capture the changes from. Must be non-empty. |
```
gunnarmorling: Is this something the user needs to provide? Or could the connector itself organize that mapping from database(s) to task(s)?

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Introduce `Partitioned*` APIs ([#13](https://github.com/sugarcrm/debezium/pull/13), [#16](https://github.com/sugarcrm/debezium/issues/16), [#18](https://github.com/sugarcrm/debezium/issues/18))
+
+####  `Partitioned*` interfaces
+
+Despite the fact that there are multiple connectors that can benefit from being aware, some connectors are nonpartitioned by their nature. For instance, the MySQL connector consumes all the changes from a nonpartitioned binlog, and it would make sense to have its API and the implementation affected by the changes being proposed as little as possible.
+
+In order to accomodate partition-aware and -unaware components under the same framework, it's proposed to introduce the `Partitioned` analogues of the following interfaces:
+
+1. `SnapshotChangeEventSource` → `PartitionedSnapshotChangeEventSource`
```
gunnarmorling: Why is it that the change event source needs to have awareness of partitioning? IIRC, during our call we assuming that the existing implementation could effectively work on multiple databases, only the configuration which the task receives needs to be adjusted?

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+Each of their methods that implements capturing data will have two more arguments added:
+
+1. `<P extends TaskPartition> taskPartition` to identify which task partition the method call should use.
+2. `<O extends OffsetContext> offsetContext` the offset context that corresponds to the partition.
+
+This way, each of the connectors could implement the relevant of the two APIs. It will also allow connectors to transition from one API to another and allow the user to chose the mode via configuration.
+
+#### Affect on framework-level classes
+
+Some framework-level classes that are reused by multiple connectors and depend on the interfaces above will have to be reworked to satisfy both of the APIs. It could be either one class per interface or one class implementing both interfaces. The latter looks more reasonable since their implementatin will be the same for the most part.
```
gunnarmorling:
```suggestion
Some framework-level classes that are reused by multiple connectors and depend on the interfaces above will have to be reworked to satisfy both of the APIs. It could be either one class per interface or one class implementing both interfaces. The latter looks more reasonable since their implementation will be the same for the most part.
```

DDD-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+So far, the following classes have been identified:
+
+1. `ChangeEventSourceCoordinator`
+2. `RelationalSnapshotChangeEventSource`
+3. `AbstractSnapshotChangeEventSource`
+
+#### Example of a change in the `ChangeEventSourceCoordinator` logic
+
+In partitioned mode, the coordinator coordinator, instead of just switching from snapshot to the streaming sources, will iterate over a collection of partitions/offsets and reuse each source for each of them:
```
gunnarmorling:
```suggestion
In partitioned mode, the coordinator, instead of just switching from snapshot to the streaming sources, will iterate over a collection of partitions/offsets and reuse each source for each of them:
```

gunnarmorling (commented):
Hey @morozov, thanks a lot! A few first comments inline. I'll let it sink in a bit and read again early next week. Should we do another call any time soon to sync on it?

DDD-1.md
morozov: Correct. The support for multiple connectors per task will allow deploying a single connector per server instance and tuning its performance by changing `maxTasks` rather than deploying a different number of connectors and manually changing the list of databases for each.

DDD-1.md
morozov: Yes. I think it could be done in the form of a "legacy" switch like it's done for the MySQL connector but with an additional opt-in step:

1. The connector implements both the old and the new APIs.
2. A new configuration parameter is added to the connector (`internal.implementation=legacy|partitioned`). The default is `legacy`. Opt-in is possible.
3. In a subsequent release, the default is changed to `partitioned`, still possible to opt-out.
4. In another release, the configuration is removed together with the legacy implementation.

**Technical details of the migration**:

Whenever a user decides to upgrade an existing connector deployment:
1. Stop the old connector (one connector = one database).
2. Consume the message with the key `["$connector_name",{"server":"$connector_name"}]` from the Kafka Connect `offsets` topic.
3. Produce a message with the same value but `["$connector_name",{"server":"$connector_name", database: "$database_name"}]`.
4. Repeat for all connectors/databases you want to migrate under one new connector.
5. Start the new connector.

It looks like only the `offsets` topic will need care. The data produced by the connector remains the same.

DDD-1.md
morozov: In our case, it's a single database. Although, the API at the framework level is generic ([`TaskPartition`](https://github.com/sugarcrm/debezium/blob/DBZ-2975/debezium-core/src/main/java/io/debezium/connector/common/TaskPartition.java)), so it allows a given connector to implement any partitioning model ([`SqlServerTaskPartition`](https://github.com/sugarcrm/debezium/blob/DBZ-2975/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerTaskPartition.java)). It should be even possible for a single connector to define different partitioning models (e.g. per-table during the snapshot).

DDD-1.md
morozov: It's the connector that will read its `database.names` and divide them between tasks (`task.database.names`). This configuration will be internal to the connector.

DDD-1.md
morozov: > IIRC, during our call we assuming that the existing implementation could effectively work on multiple databases, only the configuration which the task receives needs to be adjusted?

Yes, we did assume that. I stumbled on a few challenges during the implementation. Here's the reasoning:

1. Some connectors like the one for MySQL are nonpartitioned by their nature. Making it implement the partitioned API looks suboptimal. Both the snapshot and the streaming sources of a partition-aware connector should return to the coordinator as soon as there are no more changes in order to allow another partition to be processed. But for MySQL, there are no other partitions by definition.
2. Having two sets of APIs will allow for easier migration as the following:
   1. The connector implements the old API.
   2. The connector implements both the new and the old APIs with the ability to opt-in and opt-out.
   3. The connector stops implementing the old API.

morozov:
> Should we do another call any time soon to sync on it?

That would be cool. I'm available all week 9-10am PST. Does that work for you?

gunnarmorling:
> I'm available all week 9-10am PST. Does that work for you?

Yes; would tomorrow (Friday 19th) that time work for you?

gunnarmorling:
One semi-random note I realized when discussing this in another context: key and value schema names will need adjustment to include the database name.

morozov:
> Yes; would tomorrow (Friday 19th) that time work for you?

Yes, that works.

> One semi-random note I realized when discussing this in another context: key and value schema names will need adjustment to include the database name.

Yes, we implemented that currently w/o a migration path. Will be happy to discuss the ideas on making it smooth from the upgrade perspective.

DDD-1.md
gunnarmorling: Ok, makes sense.

DDD-1.md
gunnarmorling: Yeah, perhaps it might even be easier. For other offset format changes we did, we just made sure that the new connector version can read back the old offset format. It'd always write the new one.

DDD-1.md
gunnarmorling: Perfect.

DDD-1.md
gunnarmorling: >Both the snapshot and the streaming sources of a partition-aware connector should return to the coordinator as soon as there are no more changes in order to allow another partition to be processed.

I don't understand that part yet. Yes, the snapshot source should return once it has processed all partions, be it one, or multiple. But the streaming souce keeps running until the connector gets stopped. In my perception, the loop iterating through the databases would be _within_ the streaming source of a connector supporting multiple databases.

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+    streamingSource.execute(partition, result.getOffset())
+  })
+}
+```
+
+### Changes in mapping table identifiers to topic names and internal schema representation
+
+In order to accomodate changes from multiple databases, instead of naming topics as `<server>.<schema>.<table>`, the connector will name them as `<server>.<database>.<schema>.<table>`.
+
+The changes implemented in [DBZ-1089](https://issues.redhat.com/browse/DBZ-1089) should be partially reverted since for SQL Server both the database name and schema name should be taken into account when building schema.
```
gunnarmorling, line 183: Good point. Same for schema names, as dicussed, and some option names. In terms of migration, if a connector only captures a single database, then we may keep omitting the database name from topic names.

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+
+## TODO:
+
+1. Define configuration for multiple databases to be processed by a connector. Only one of `database.dbname` and this one must be configured on a given connector.
+2. Provide the upgrade path. The addition of the database name to the records' source partition will make it impossible for the upgraded connector to resume from the offset produced by the previous version. Likely, we'll need to keep the "legacy" version and let the consumers migrate their data and switch over.
```
gunnarmorling: Why do you think would it be impossible for the connector to read back the old offset format?

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+1. Describe a *set* of source partitions to be processed by the task.
+2. Initialize the set above. Normally, based on the task configuration.
+3. Represent a source partition in higher-level terms specific to a given connector (e.g. database name).
+4. Allow mapping of connector-level partition (e.g. a database name) to the corresponding Kafka Connect source partition.
+5. Allow mapping of connector-level partition to the corresponding `OffsetContext`.
+6. Instead of a single `OffsetContext` per task, use a map of connector-level partitions to their offset context.
+
+### Support for multiple tasks per connector ([#9](https://github.com/sugarcrm/debezium/issues/9))
+
+Apart from being able to process multiple databases by a given task, it would be great if a connector was able to start multiple tasks and divide the databases between them. This way, it would allow to scale a single connector just by changing its `max.tasks` configuration.
```
gunnarmorling: Would this apply to snapshotthing _and_ stream-reading, or only the former?

gunnarmorling: Or, asked differently, would it make sense with a simplified model, where the connector supports exactly one task per database?

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+
+| Name                  | Type     | Default | Importance | Doc                                                          |
+| :-------------------- | :------- | :------ | :--------- | :----------------------------------------------------------- |
+| `task.database.names` | `STRING` | `None`  | `HIGH`     | The list of comma-separated database names form which the given connector task should capture the changes from. Must be non-empty. |
+
+The connector can monitor the list of CDC-enabled databases and maintain the list of databases to capture the changes from ([#11](https://github.com/sugarcrm/debezium/issues/11)). This will require another configuration parameter:
+
+| Name            | Type      | Default | Importance | Doc                                                          |
+| :-------------- | :-------- | :------ | :--------- | :----------------------------------------------------------- |
+| `to.be.defined` | `BOOLEAN` | `False` | `Medium`   | If set to true, the connector will monitor the list of CDC-enabled databases and capture changes from all of them. |
```
gunnarmorling: So this would be mutually exclusive with the existing `database.dbname`, right? I'm wondering whether we should simply allow to give multiple names using that option (or have a new one, `dbnames`)? Could there be a case where one would not like to capture _all_ CDC enabled databases, but only a sub-set? I fell this level of control might be desirable.

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Recover database schema from multiple partitions in one pass ([#14](https://github.com/sugarcrm/debezium/issues/14))
+
+Currently the internal schema representation used by the connector is capable of representing multiple databases (e.g. for the MySQL connector) but isn't capable of initializing from the database history of multiple partitions. It should be possible to initialize it multiple times, once for each partition but it may be suboptimal, especially if the database history topic and the number of partitions are large.
```
jpechane: But is it capable of mutlithreaded access?

DDD-1.md
```diff
@@ -0,0 +1,179 @@
...
+
+Instead, the schema could recover from multiple offsets in one pass.
```
gunnarmorling: The alternative would be to have a separate history topic per task/partition. Just throwing out the idea, not saying it's necessarily better.

DDD-1.md
gunnarmorling: Doesn't make sense due to connection usage for that. Thanks!

gunnarmorling:
Open question: how to deal with metrics?

DDD-1.md
morozov: I assume, this question applies regardless of whether we want to initialize the schema for all partitions at once or sequentially. Even if we initialize the schema sequentially for each partition, multiple tasks running on the same worker may still access it in parallel. We'll have to check that additionally.

DDD-1.md
morozov: From the logic perspective, it's a minor change: primarily needs changes in the comparator that determines the stop point (https://github.com/sugarcrm/debezium/pull/26).

DDD-1.md
morozov: We'll need to do some more research here. At the moment I'm not sure that the "CDC-enabled" attribute is a reliable signal for the connector to start capturing changes from the database. If it was, then on top of that flag, a set of `database.include.list`/`database.exclude.list` could be added similarly to the MySQL connector.

DDD-1.md
morozov: No reason to think it won't work. I like the approach.

morozov:
I looked a bit deeper at using the [Signaling Table](https://debezium.io/blog/2021/03/15/debezium-1-5-beta2-released/) API to implement snapshotting new databases by the same connector that captures streaming changes from the same server instance.

If I understood the idea correctly, the mechanics would the following:

1. The database server sends a "snapshot" command containing the database name to the connector.
2. The connector starts a new task to snapshot the new database. Looks like the `max.tasks` config parameter is indeed [advisory](https://github.com/apache/kafka/blob/a63e5be4195e97e5b825b5912291144d2d0283a3/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/Worker.java#L351-L373). The code doesn't prevent the connector from returning more configurations than `max.tasks`.
3. The task performs the snapshot w/o blocking the rest of the streaming tasks.
4. Once the snapshot is done, the connector assigns the database to one of its tasks.

There are a few unclear moments:

1. The changes in the signaling table will be processed by one of the connector *tasks*. While it's the *connector* that will need to schedule a new snapshotting task. There doesn't seem to be away for the task to communicate back to the connector (at least judging by the `SourceTaskContext` API).
2. Unless the task can communicate back to the connector, there's no clear way to determine the snapshot completion at the connector level.
3. There is a principal difference between, for example, the `SchemaChanges` command implemented via the signaling table, and the `Snapshot` command. The result of processing the `SchemaChange` command is produced to one of the topics produced by the connector before processing the following command. So if the connector commits its offset, there is a guarantee that the command has been fully processed. This doesn't apply to the `Snapshot` command since it's meant to be performed asynchronously to capturing CDC events. At the same time, there's no way for the connector to persist its configuration changes.
4. Once the snapshot is done, the connector will need to assign the database to one of its streaming tasks. This is possible at runtime but it won't be persisted in the connector configuration (i.e. the `database.names` stored in Kafka Connect will remain the same until changed via the API).

Technically, most of the concerns could be addressed by adding a "signaling" topic through which the task could interact with the connector, it could also store the state of snapshots. But I'm not sure if the framework will benefit from that. Maybe the same is possible via the existing Kafka Connect topics (assuming they are supposed to be used by the implementations).

Did I get the idea correctly? What do you think?

jpechane:
@morozov
1. In this regard IIRC the proposal was to have one task parked for exactly hadnling such situations. While you cannot communicate with connector itself we can add a way for tasks to communicate among themselves
2. That's IMHO unnecessary
3. It is not there yet but I envisioned exactly this problem. The idea is to store current running command in the offsets so it will resume upon restart. In the same way thare are tx specific data there will be signal/action specific data stored in offsets
4. Aain this could be solved via offsets. The offsets could contain extra list of database names that would be used for filtering and cleared when those database names will be stored in the config.

morozov:
@jpechane in your comment above, do you refer to the first list or to the second one? Both have 4 points in them. Judging by #​3 in your response, it's about the second list.

> In this regard IIRC the proposal was to have one task parked for exactly hadnling such situations.

I must have missed this bit. In this case, as I understand, this task will read the changes from the database where the signaling table resides and do the snapshots. Right? This way, the connector could do only one snapshot at a time which isn't a problem per se but seems to conflict with some statements below.

> That's IMHO unnecessary

Makes sense if I understand the above right.

> The idea is to store current running command in the offsets so it will resume upon restart.

What database should these offsets belong to? It's either the database where the signaling table resides or the database to be snapshotted. I assume it's the former because if the connector restarts it won't even know that it needs to consume the offsets of the database to be snapshotted.

In this case, as far as I understand, the task that captures the changes from the signaling table will have to produce some source records from which Kafka Connect would derive and store the offsets. To which topic should those messages be produced? Or maybe a task can commit its offsets directly w/o producing source messages?

> The offsets could contain extra list of database names that would be used for filtering

I assume it's expected that the connector is configured with a full list of databases including those that haven't been snapshotted yet. The streaming tasks will ignore the databases that haven't been snapshotted until their snapshots are done. Does it mean that the streaming tasks should consume the offsets topic?

A "list of database names" implies that there may be multiple snapshots being done in parallel. Is it possible if the parked task will read the signaling table and do the snapshots, one at a time?

jpechane:
@morozov Hi, yes these were answers to the unclear points

> I must have missed this bit. In this case, as I understand, this task will read the changes from the database where the signaling table resides and do the snapshots. Right? This way, the connector could do only one snapshot at a time which isn't a problem per se but seems to conflict with some statements below.

Well, IIRC that was on of the proposals how to solve the problem that it is not possible to start the taskas needed. @gunnarmorling then proposed to have a back task parked to handle this cases. And yes you are right, in this case it will execute the snapshot serially.

> What database should these offsets belong to? It's either the database where the signaling table resides or the database to be snapshotted. I assume it's the former because if the connector restarts it won't even know that it needs to consume the offsets of the database to be snapshotted.

I understand this is a little bit unclear right now but I believe it should go to both databases. For the signalling table database it is necessary to be stored there to highlight the action has alread been started. For the new database it is exactly for the purpose of having the conector to know that the database is already captured.

> In this case, as far as I understand, the task that captures the changes from the signaling table will have to produce some source records from which Kafka Connect would derive and store the offsets. To which topic should those messages be produced? Or maybe a task can commit its offsets directly w/o producing source messages?

Well, in case of high-traffic database the offsets could delivered as part of regular data message. But you are right, this is not guranteed so we might need to introduce a separate topic to maybe even send some status messages or reuse heartbeat topic for that?

> I assume it's expected that the connector is configured with a full list of databases including those that haven't been snapshotted yet. The streaming tasks will ignore the databases that haven't been snapshotted until their snapshots are done. Does it mean that the streaming tasks should consume the offsets topic?

The streaming tasks will receive thier offsets via the offset loading API call. Win't they?

> A "list of database names" implies that there may be multiple snapshots being done in parallel. Is it possible if the parked task will read the signaling table and do the snapshots, one at a time?

Sorry for the confusion, it was not intended in that way. It was intended as a queue of databases not yet snapshotted and they'd be snapshotted one by one as mentioned above.

morozov:
@jpechane thank you for the clarification. I think I understand the idea at a high level, and it looks like it's farther beyond the existing Debezium architecture than my initial proposal about multi-tasking/multi-partitioning. At least given my current knowledge of Debezium.

If it's okay with you and Gunnar, I'd like to put this part of the proposal on hold and keep focusing on the multi-tasking/multi-partitioning support. In the meanwhile, we will temporarily address the scenario of adding a new database by a semi-automated scripted solution via a temporary snapshotting connector. I'll be happy to document the details if you're interested. And yes, you were right: the stored offset unconditionally contains the connector name :-)

> The streaming tasks will receive their offsets via the offset loading API call. Won't they?

They will receive all relevant offsets once during the start, yes. But if we want to hand over a snapshotted database from the snapshotting task to the streaming one, something (maybe the connector) should consume the offsets topic and wait until the last snapshotted message is produced. Or poll the `OffsetStorageReader` (no idea how efficient that would be).

> […] we might need to introduce a separate topic to maybe even send some status messages or reuse heartbeat topic for that?

Yeah, I wasn't aware of the heartbeat feature but it looks relevant to what we'd need.

gunnarmorling:
> In the meanwhile, we will temporarily address the scenario of adding a new database by a semi-automated scripted solution via a temporary snapshotting connector. I'll be happy to document the details if you're interested.

I'm definitely interested :) In particular, how do you intend to implement the switch from that snapshotting connector over to the regular connector?

DDD-1.md
gunnarmorling: Ok, cool. Let's do this then.

morozov:
Here's the full process including the switch over (starting step 4).

A few notes:
1. We use Strimzi to manage Kafka resources in Kubernetes, so the connector specs are in YAML.
2. We route all messages from each database to a single topic (topic named equals to the DB name).

Assuming there is an already running connector (hereinafter 1:M connector) that captures changes from the databases "a" and "b" on a given server:

```yaml
metadata:
  name: sql-01
spec:
  database.server.name: sql-01.example.com
  database.names: a,b
  database.history.kafka.topic: sql-01.dbhistory
```

A new database "c" needs to be added to the connector.

1. Deploy a temporary snapshot connector:

   ```yaml
   metadata:
     name: sql-01-c-snapshot
   spec:
     database.server.name: sql-01.example.com
     database.names: c
     snapshot.mode: initial_only
     database.history.kafka.topic: c.snapshot.dbhistory
   ```

   It's safer to use a dedicated database history topic to be able to start from scratch if the snapshot fails. The rest of the configuration should be the same as in the 1:M connector. Maybe there's a better solution.

2. Wait until the snapshot is done:

   1. Start consuming from the end of the Kafka Connect offsets topic using `kafka-console-consumer`.

   2. Consider only the messages that belong to the snapshot connector. Their key should look like the following (formatted for readability):

      ```yaml
      [
          "sql-01-c-snapshot",
          {
              "server": "sql-01",
              "database": "c"
          }
      ]
      ```

   3. If the offset has `{"snapshot": "last"}`, break. Otherwise, continue consuming.

3. Delete the snapshot connector.

4. Copy database history from the temporary topic to the primary one without any modification (`kafka-console-consumer -t c.snapshot.dbhistory | kafka-console-producer sql-01.dbhistory`).

5. Use the value of the last message consumed on step 2 and produce a message to the Kafka Connect offsets topic with the same value and the key of the 1:M connector (formatted for readability):

   ```yaml
   [
       "sql-01",
       {
           "server": "sql-01",
           "database": "c"
       }
   ]
   ```

6. Update the list of database names on the running connector:

   ```yaml
   spec:
     database.names: a,b,c
   ```

7. Delete the `c.snapshot.dbhistory` topic.

Upon restart, the connector should pick up from where the snapshot connector left off.

DDD-1.md
gunnarmorling: @morozov, so I believe I understand now, that you envision to have one streaming change event source _per partition_ (i.e. database), correct? So if e.g. a task processes three partitions, there'd be three streaming change event sources, and the coordinator cycles through them. Is that what you have in mind? If so, I don't get yet the change of the pending PR, where offsets are moved to method parameters. Would this be actually necessary in this case? I feel I'm still misunderstanding something here.

gunnarmorling:
@morozov, hey there, completely missed that this one still is pending. Could you perhaps take a look into this and see whether there are any open discussion items inline and try to wrap up those one? Would be nice to have this merged, so we have a referenceable description of the intended (and hopefully actual) implementation of this feature. Thanks a lot, and sorry again for the long delay here.

DDD-1.md
morozov: This is outdated. I removed this section.

DDD-1.md
morozov: As discussed offline, the SQL Server task will process each partition sequentially, so there is no need for the schema to be capable of handling multithreaded access.

I'm not familiar with how schema handling works in the MongoDB connector, so we may need to revisit this question if/when this approach is applied to the MongoDB connector.

morozov:
@gunnarmorling please see the update.

morozov:
@gunnarmorling do you want to merge this document in its current state? It's primarily up to date except for the "metrics" part. The only mismatch is that instead of reworking the core class hierarchy, we introduced the `Partition` arguments to the listeners and implemented a separate hierarchy for SQL Server.

I'm afraid, describing the design in a text format will take quite some time, and hopefully, the code speaks better for itself.

What do you think?

gunnarmorling:
@morozov, yeah, let's merge it as-is, I think it definitely provides value in the current form. Thanks a lot!

PR #6, opened 2023-04-17
DDD-4 Offset Read/Write API (Core/UI)

Naros:
A design document outlining the introduction of changes for the https://github.com/debezium/debezium repository and the https://github.com/debezium/debezium-ui repository to support that reading and writing of offset details for connectors.

### DIFF
--- a/DDD-4.md
+++ b/DDD-4.md
@@ -1,0 +1,348 @@
+# DDD-4: Offset Read/Write API
+
+## Motivation
+
+Debezium is a platform to stream changes executed on data by processing transaction logs of a database.
+As changes are processed, a position into the transaction log is maintained so that Debezium can safely resume on restarts.
+
+This position, often referred to as offsets, differs by connector.
+While the MySQL connector maintains a reference to the binlog filename and a position within the binlog, other connectors maintain state that is specific to that database vendor's implementation.
+Therefore, there is no single common structure for these offsets.
+
+Debezium UI is a web-based application that is meant to simplify the deployment and management of Debezium connectors.
+We want to improve the user experience with the UI by introducing the ability to perform the following tasks:
+
+* Start, Stop, Pause, and Resume Incremental Snapshots
+* Manipulate connector offsets
+
+Both of these tasks require the ability to read connector offsets and the latter requires the ability to write to the offsets.
+
+This document is currently based on **Debezium 2.2.0.CR1**.
+
+## Goals
+
+* Provide support for reading connector offsets
+* Provide support for writing connector offsets
+* Provide a common schema-based API for working with offset data that is universal for all connectors
+* Integrate this behavior into the Debezium UI backend
+
+## Current connector offset behavior
+
+Debezium 2.x supports the notion of partition-based offsets based on the following contracts:
+
+* `io.debezium.pipeline.spi.OffsetContext`
+* `io.debezium.pipeline.spi.OffsetContext$Loader`
+* `io.debezium.pipeline.spi.Partition`
+* `io.debezium.pipeline.spi.Partition$Provider`
+* `io.debezium.connector.common.OffsetReader`
+
+These contracts allow the reading of the offset metadata by providing an instance of `org.apache.kafka.connect.storage.OffsetStorageReader` via the `org.apache.kafka.connect.source.SourceTaskContext`, enabling the connector to deserialize the offsets from the appropriate storage implementation.  
+
+## Current connector offset attributes
+
+The following describes the current offset data managed by each connector implementation.
+
+### Cassandra 
+
+The Cassandra 4 and 4 implementations maintain identical offset formats.
+The Cassandra connector maintains a key/value pair map that consists of keys based on the keyspace table name and values that refer to a serialized string using the format of `<file-name>:<file-position>`.
+
+So for example, assuming a keyspace `cycling` with a table named `birthday_list`, if a change was detected for this table at position `12345` in the commit log, a entry would exist in this map similar to the following:
+
+```json
+{ "birthday_list" : "commitlog-<version>-<msec_timestamp>.log:12345" }
+```
+
+For Cassandra, we need to support:
+
+* Adding or Removing keys from this map.
+* Modifying the value associated with a given key in this map.
+
+### Db2
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,3 +8,4 @@ When creating a new document, please add it to the list below.
 
 * [DDD-1](DDD-1.md): Support for multiple source partitions for relational Debezium connectors
 * [DDD-3](DDD-3.md): Incremental snapshotting
+* [DDD-4](DDD-4.md): Oracle Read/Write API
```
jpechane, line 11:
```suggestion
* [DDD-4](DDD-4.md): Offset Read/Write API
```

DDD-4.md
```diff
@@ -0,0 +1,348 @@
+# DDD-4: Offset Read/Write API
```
jpechane, line 1: Do I understand it correctly that we are talking about **REST** API? I suppose from the context but it should be highlighted in the naming as well.

DDD-4.md
```diff
@@ -0,0 +1,348 @@
...
+While the MySQL connector maintains a reference to the binlog filename and a position within the binlog, other connectors maintain state that is specific to that database vendor's implementation.
+Therefore, there is no single common structure for these offsets.
+
+Debezium UI is a web-based application that is meant to simplify the deployment and management of Debezium connectors.
+We want to improve the user experience with the UI by introducing the ability to perform the following tasks:
+
+* Start, Stop, Pause, and Resume Incremental Snapshots
+* Manipulate connector offsets
+
+Both of these tasks require the ability to read connector offsets and the latter requires the ability to write to the offsets.
```
jpechane, line 18: Why are they required by incremental snapshots? To get the state and see if it is running?

DDD-4.md
```diff
@@ -0,0 +1,348 @@
...
+
+These contracts allow the reading of the offset metadata by providing an instance of `org.apache.kafka.connect.storage.OffsetStorageReader` via the `org.apache.kafka.connect.source.SourceTaskContext`, enabling the connector to deserialize the offsets from the appropriate storage implementation.  
+
+## Current connector offset attributes
+
+The following describes the current offset data managed by each connector implementation.
+
+### Cassandra 
+
+The Cassandra 4 and 4 implementations maintain identical offset formats.
```
jpechane, line 47:
```suggestion
The Cassandra 3 and 4 implementations maintain identical offset formats.
```

DDD-4.md
```diff
@@ -0,0 +1,348 @@
...
+The `commit_scn` field is required by the Oracle connector when using the LogMiner adapter (the default).
+This field has undergone a significant transformation over the past several releases and therefore its critical to understand how to serialize/deserialize this as indicated below:
+
+| Version           | Format                                                                     |
+|-------------------|----------------------------------------------------------------------------|
+| Up to 1.9.4.Final | `<commit-scn>`                                                             |
+| 1.9.5.Final       | `<commit-scn>:<rollback-segment-id>:<sql-sequence-number>:<arch-thread-#>` |
+| 1.9.6.Final       | `<commit-scn>:<arch-thread-#>:<transaction-id-list>`                       |
+| 1.9.7.Final       | `<commit-scn>:<arch-thread-#>:<transaction-id-list>`                       |  
+| 2.0.0.Final+      | `<commit-scn>:<arch-thread-#>:<transaction-id-list>`                       |
```
jpechane, line 181: I suppose we are only going to support only 2.x offsets.

jpechane (commented):
The was a plan for 2.0 (not done unfortunately) to add a versioning support to offsets to enable easy offset fromat change/upgrades. Do you think it would be worthy to get back to it side-by-side with this?

DDD-4.md
Naros: There is a REST API aspect to this, which is yet to be documented, but I also was going through each connector and looking at the offsets and what attributes each possessed.  

We need to have a way to pass offset fields with their name, type, and value to the front end, be able to define any relationship between fields or at the very least validate UI input against such relationships, and then have a way where we read/write the offset data to the offset backing store.

As a part of this effort, I feel like we should have some implementation detail in each connector that we can leverage as a part of this REST API, allowing the bulk of the logic for each connector's validation or marshaling to be a detail of the connector itself, not the API needed for the REST extension and/or UI backend.

As for the name, we can certainly find a better naming, I'm open to suggestions.

DDD-4.md
Naros: Correct, we don't want to allow the user to stop an incremental snapshot if one isn't currently running.  While we do understand that sending such a signal is ignored (no-op), it's so we design a good UX.

DDD-4.md
Naros: Most likely, but I added this mainly for completeness.

Naros:
> The was a plan for 2.0 (not done unfortunately) to add a versioning support to offsets to enable easy offset fromat change/upgrades. Do you think it would be worthy to get back to it side-by-side with this?

It may be worth it.  

As I mentioned earlier, we need to have a way in which offset structures are defined to be consumed in a way that allows the UI to provide similar UX widget experiences like what we have for other connector configuration properties.  Whether we have a similar `ConfigDef` type object for this or something less polished, at a minimum, the UI needs a way to unmarshal offsets from the backing store into JSON with details about each field, perhaps a description, name, type, and then a way to verify the user input and marshal the data for the backing store.

When we look at connectors like Oracle, there are fields in the offsets that should only be applicable based on connector configuration, i.e. `lcr_position` shouldn't exist if the connector uses LogMiner.

jpechane:
@Naros Could we the connect schema to describe the offsets?

Naros:
> @Naros Could we the connect schema to describe the offsets?

I think we can in part, the only question is how to deal with some of the situations where we serialize a field as a STRING and yet the field's contents have some internal representation, such as Oracle's `commit_scn` field.  

Perhaps we should consider adopting a fully structured implementation for the duration of Debezium 2.x where we serialize both such fields as their STRING representation and a structured one?

jpechane:
Is this DDD affected by https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect ?

PR #7, opened 2023-05-23
DDD-6 Debezium Operator

jcechace:
First draft of DDD for Debezium Operator. This should be iteratively improved and serve as a preliminary documentation.

### DIFF
--- a/DDD-6.md
+++ b/DDD-6.md
@@ -36,65 +36,139 @@ Initially will start with support for Debezium Server. However the resources sho
 The two primary options to consider her are Golang's Operator SDK (Golang) or Java Operator Framework (as Quarkus extension) . In case of a java-centered team such as our it makes sense to go with JODSK --  especially since JOSDK has quite active development. 
 
 
-### Custom Resource Proposal
-The following is a proposed example of `DebeziumServer` custom resource instance (in this). 
+## Custom Resource Proposal
+
+### DebeziumServerSpec Reference
+```yaml
+spec:
+  version: String
+  image: String # exclusive with version
+  storage:
+    type: persistent | ephemeral  # enum
+    claimName: String # only valid and required for persistent
+  runtime:
+    env: EnvFromSource array # https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envfromsource-v1-core
+    volumes: Volume array # https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#volume-v1-core
+  quarkus:
+  # quarkus properties 
+  format:
+    value:
+      type: String
+      # other format properties
+    key:
+      type: String
+    header:
+      type: String
+  transforms:
+    - type: String
+      predicate: String
+      negate: Boolean
+      config:
+        # other transformation properties
+  predicates:
+    name:
+      type: String
+      config:
+        # other preticate properties
+  sink:
+    type: String
+    config:
+      # other sink properties
+  source:
+    class: String
+    config:
+      # other source connector properties
+```
+
+Resource status will be an array of `Condition`. 
+### DebeziumServerStatus Reference
+```yaml
+status:
+    conditions: # Condition array
+        - status: String
+          type: String
+          message: String
+
+```
+For start only the `Ready` condition is defined
+```
+status: True | False
+type: Ready
+```
+
+### Example Custom Resource Instance
+The following is an of the proprosed `DebeziumServer` custom resource. 
 
 ```yaml
 apiVersion: debezium.io/v1alpha1
 kind: DebeziumServer
-metadata: 
-    name: my-debezium
+metadata:
+  name: my-debezium
 spec:
-    version: "2.2"
-    storage:
-      type: persistent 
-      claimName: ds-data-pvc
-    quarkus:
-      log.console.json: false
-    format:
-      value:
-        type: json
+  image: quay.io/jcechace/debezium-server:latest
+  storage:
+    type: persistent
+    claimName: ds-data-pvc
+  runtime:
+    env:
+      - configMapRef:
+          name: ds-env
+      - secretRef:
+          name: ds-secret-env
+    volumes:
+      - name: extra
+        configMap:
+          name: ds-mounts
+      - name: extra-secret
+        secret:
+          secretName: ds-secret-mounts
+  quarkus:
+    log.console.json: false
+    kubernetes-config.enabled: true # enable access to config maps and secrets
+    kubernetes-config.secrets: ds-creds # use ds-creds secret in the same namespace
+  format:
+    value:
+      type: json
+      config:
         schemas.enable: false
-    transforms:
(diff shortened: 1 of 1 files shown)


DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+    conditions:
+        Status: True
+        Type: Ready
+        Message: Server my-debezium is operational 
+```
+
+
+## Open Questions
+
+1) Do we need a service? What for?
```
jpechane: I don't think we need service right now. Debezium does not expose any data endpoint.

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+        type: json
+        schemas.enable: false
+    transforms:
+      - name: test
+        type: com.example.TestTransform
+        predicate: test
+        negate: false
+        prop: 42
+    sink:
+      type: kafka
```
jpechane: I suppose `type` is just plain string, not enum of allow sinks, is it correct?

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+      producer.bootstrap.servers: dbz-kafka-kafka-bootstrap:9092
+      producer.key.serializer: org.apache.kafka.common.serialization.StringSerializer
+      producer.value.serializer: org.apache.kafka.common.serialization.StringSerializer
+    source:
+      class: io.debezium.connector.mongodb.MongoDbConnector
```
jpechane: Probably not now byt we might consider to use a logical name instead of a class name.

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+      mongodb.connection.string: mongodb://debezium:dbz@mongo.debezium.svc.cluster.local:27017/?replicaSet=rs0
+      database.history: io.debezium.relational.history.FileDatabaseHistory
```
jpechane: Probably not now byt we might consider to use a logical name instead of a class name.

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+
+## Open Questions
+
+1) Do we need a service? What for?
+2) How to handle "Templating" to allow customization of deplyments outside of the scope allowed by the operator 
+    - This is needed due to reconciliation (JOSDK doesn't allow merge reconciliations)
+    -  It also solves a problem of different requirements between sinks (e.g. only some require environment variables)
+3) Dynamic vs Static cnfiguration typing
```
jpechane: Let's start with the dynamic approach. It is the easiest to implement.
To de-couple sink/operator development we might allow combination of both. Does it make sense to introduce a separate namespace already that will represent the dynamic configuration and will be the only option right now while later on we could introduce statics too?

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+    - Per sink type configuration with specific properties
+    - Single sink configurationion with arbitrary properties in form of `Map<String, Object>`
+4) Per namespace / all namespaces / both?
```
jpechane: I'd start with per-namespace as it is least compplicated from k8s - speaking of security and privileges.

DDD-6.md
```diff
@@ -0,0 +1,100 @@
...
+        Status: True
+        Type: Ready
+        Message: Server my-debezium is operational 
+```
+
+
+## Open Questions
+
+1) Do we need a service? What for?
+2) How to handle "Templating" to allow customization of deplyments outside of the scope allowed by the operator
```
jpechane: I'm sorry but I don't follow this question. Could you please share an example?

jcechace: Essentially an equivalent of https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaClusterTemplate-reference

DDD-6.md
jcechace: I'm not really sure what you mean by namespace in this context. The idea here is that the properties under sink are mostly just gathered into a map. 

In theory in future we could do something such as 

```yaml
sink:
  kafka:
    bootstraps: kafka:9092
```

so esentially we would introduce a typed property for each sink

DDD-6.md
jcechace: Yes, for the "typed" configuration this would make sense.

DDD-6.md
jcechace: yes, but changing it to enum would be quite simple. However an "outdated" operator would be unusable in case of a new sink release.

DDD-6.md
jcechace: +1 I will remove the service then

DDD-6.md
jpechane: Yes, I understand, my plan was to hve something like
```
sink:
  kafka:
    bootstraps: kafka:9092
    additional:
       prop1:
       prop2:
```
So you could use arbitrary props even with typed properties in place - for cases when the is a delay between operator/server, or if something is omitted.
Namespace in this case is the separate field - here named `additional` that contains this dynamic configuration.
What I was talking about was placing the conf options under such distinct field so it will stay bacward compatible when typed configuration will be available.

DDD-6.md
jpechane: I personally prefer the plain string - for sake of flexibility.

DDD-6.md
jcechace: Yes, this can be there for sure. The question is what should be the initial state before we introduce these typed sink configurations. The way I see it we have three option

1) Current version 

```yaml
sink:
  type: kafka
  prop1:
  prop2: 
```

then later we could add the typed versions

```yaml
sink:
  kafka:
    bootstraps: kafka:9092
    additional: 
      prop1:
      prop2:
```  

in this case the `type` and additional props would still be directly part of `Sink` but validation would be put in place to ensure these configs are mutually exclusive

2) Slightly modified current verison

```yaml
sink:
  type: kafka
  additional: 
    prop1:
    prop2:
``` 

Then again, after introducing the typed `kafka` (and so on) config we would ensure that `type` is mutually exclusive (sink.additonal could be then merged into sink.kafka.additional)

3) We initially introduce an arbitrary sink type 
```yaml
sink:
  config:
    type: kafka
    bootstraps: kafka:9092
    prop1:
    prop2:
```  
In this scenario `sink.config` is essentially an extra nesting of current `sink`.

jcechace: Or... thinking about it.. which is probably the best from implementation point of view. 

The initial version is going to be 

```yaml
sink:
  type: kafka
  config:
     prop1:
     prop2:
```

Once again, `config` is essentially an arbitrary dynamic sink type (and also technical equivalent of the `additional` filed)

And then later once typed configs are introduced

```yaml
sink:
  type: kafka # must be kafka for kafkaConfig
  kafkaConfig:
    bootstraps: kafka:9092
  config: 
    prop1:
    prop2:
```
This gives as the best backwards compatibility and also allows us to use the `type` field as a decision point for which typed config to  read (if none of known values matched, then the dynamic `config` would be expected)

WDYT?

PR #8, opened 2023-11-16
DDD-7 Asynchronous Debezium Embedded Engine

vjuranek:
https://issues.redhat.com/browse/DBZ-7073

### DIFF
--- a/DDD-7.md
+++ b/DDD-7.md
@@ -1,0 +1,153 @@
+# DDD-7: Asynchronous Debezium Embedded Engine
+
+## Motivation
+
+Debezium comes in two basic flavors.
+One flavor is  [Kafka Connect](https://docs.confluent.io/platform/current/connect/index.html) source connectors.
+The other is an independent engine, which can be embedded into a user application or is wrapped in the [Debezium server](https://debezium.io/documentation/reference/2.4/operations/debezium-server.html), which is a standalone application.
+While in the first case the connector life cycle and execution of connector tasks are managed by Kafka Connect, in the later case the task life cycle is managed purely by the Debezium engine itself and the Debezium project has full control how the tasks within the engine are executed.
+Current implementation of the [`DebeziumEngine` interface](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java), [`EmebeddedEngine`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java#L91), executes all the steps in a serial manner.
+This includes executing event transformations as well as event serialization.
+Event serialization is not provided directly by `EmebddedEngine`.
+It's provided by its extension, [`ConvertingEngine`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java#L169), which implements it as part of [event processing](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java#L102).
+
+Moreover, current `EmebddedEngine` implementation doesn't support executing multiple source tasks, even in case the source connector supports it, such e.g. [SQL server connector](https://debezium.io/documentation/reference/2.4/connectors/sqlserver.html#sqlserver-property-tasks-max).
+Only the [first task is executed](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java#L918), no matter how many tasks connector configuration provides.
+
+In the era of large data sets and multicore servers, using a single thread for processing all the CDC events from the database is a clear limiting factor.
+Providing a new implementation of `DebeziumEngine` interface which would run some of the tasks in parallel can give significant performance boost.
+New implementation and related changes should also aim for a good test coverage and ease of testing of the new implementation as well as any other future implementations.
+
+## Goals
+
+Provide new implementation of [`DebeziumEngine` interface](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java) which would
+
+* allow to run multiple [source tasks](https://github.com/apache/kafka/blob/3.6.0/connect/api/src/main/java/org/apache/kafka/connect/source/SourceTask.java) for given connector if the connector provides multiple tasks
+* run potentially time-consuming code (e.g. event transformation or serialization) in the dedicated threads
+* allow possible further speedup by optionally disabling total ordering of the messages
+* be well-prepared for future changes and new features, namely
+  * switching to [virtual threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html)
+  * eventually delegate source tasks to external workers via [gRPC](https://grpc.io/) calls (especially to be able to horizontally scale connector load to multiple pods in Kubernetes cluster)
+  * better integration with Debezium k8s operator and UI
+  * providing Debezium engine as a [Quarkus](https://quarkus.io/) extension
+
+Another high-level goal is to adjust the current Debezium testsuite to use the `DebeziumEngine` interface instead of hard-coded `EmebddedEngine` implementation.
+This should allow for easy switching to any other implementation of `DebeziumEngine` and thus making new implementation to be easily tested with the current testsuite.
+
+## Non-goals
+
+* Change [`DebeziumEngine`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java) interface.
+* Do any other changes in the Debezium [API package](https://github.com/debezium/debezium/tree/v2.4.0.Final/debezium-api).
+* Implement any parallelization inside connectors (e.g. tracking database CDC changes in multiple threads).
+* Remove dependency on Kafka Connect API.
+
+### Preserving Kafka Connect model
+
+As the main goal of the Debezium engine is to be able to execute Debezium outside of Kafka, it may be strange why not to take this opportunity to get rid of Kafka dependencies.
+The reason is simple: it would be a too complex change which would not impact only the Debezium engine.
+E.g. removing [`WorkerConfig`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConfig.java) would require removing [`OffsetBackingStore`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetBackingStore.java),
+which would require removing [`OffsetStorageReader`](https://github.com/apache/kafka/blob/3.6.0/connect/api/src/main/java/org/apache/kafka/connect/storage/OffsetStorageReader.java) etc., etc., resulting in substantial changes in the Debezium core and connectors.
+Therefore, this should be done in a separate task which would deserve a dedicated DDD describing all the changes and proposing replacements for Kafka Connect interfaces and classes.
+
+## Proposed changes
+
+### Thread pools
+
+Implementation of concurrent processing will be based on the Java [Executors](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html) framework.
+As the Debezium is currently based on Java 11, new concurrency features introduced by the project Loom, namely virtual threads and structured concurrency, cannot be used now.
+However, it's expected to switch to virtual threads in the future, once Debezium will be based on Java 21 or higher.
+[`ThreadPoolExecutor`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html) will be used for the creating and pooling of the threads.
+The `ThreadPoolExecutor` will be created by factory method [`Executors.newFixedThreadPool​(int nThreads)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html#newFixedThreadPool(int)).
+Number of worker threads will be specified either by the user via configuration or the default value will be used.
+The default value will be the number of the cores of the underlying machine.
+There will be only one thread pool, the one for processing CDC event pipelines.
+`SourceTasks`s, which will be also run in separate threads, start and stop with the engine and therefore don't need any thread pooling.
+
+### Running source tasks in parallel
+
+Kafka Connect `SourceTask` life cycle will be separated into a self-contained task.
+These tasks will be executed in dedicated threads created by [`Executors.newSingleThreadExecutor()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html#newSingleThreadExecutor()).
+In case of `RetriableException` the task will be restarted.
+In other cases, all other tasks will be stopped gracefully and the engine will fail with the task exception.
+
+### Processing CDC events concurrently
+
+Current `DebeziumEngine` API partially delegates event processing to the [`ChangeConsumer`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java#L159) implementation, which is provided by the user.
+More specifically, events are passed to the [`ChangeConsumer#handleBatch()`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java#L167) in a batch manner, as a `List`.
+This makes sense as in many cases events are submitted to another system and this is usually more efficient to do in a batch.
+However, this prevents us from making a complete event processing pipeline, which would be run in a dedicated thread.
+
+Before passing a batch of events to the `ChangeConsumer`, user defined single message transformations are applied to the records.
+In the case of [`ConvertingEngineBuilder`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java#L53), events are moreover serialized to `SourceRecords` before passing to the `ChangeConsumer`.
+Both these tasks will be run in parallel in `ThreadPoolExecutor` mentioned above.
+Before passing the batch to the `ChangeConsumer`, tasks for all records in the batch will be awaited to finish.
+
+The implementation should also try to optimize record serialization/deserialization in case of the `ConvertingEngineBuilder`.
+Currently, the records are serialized to `SourceRecords` before passing to the `ChangeConsumer` and then deserialized back when `ChangeConsumer` calls `RecordCommitter`.
+TBD task is to find a way how to avoid record deserialization during commit without breaking existing API.
+
+#### Disabling total ordering of the CDC events
+
+Performance-wise, further speedup can be achieved by skipping message ordering and delivering messages in order they are ready to go.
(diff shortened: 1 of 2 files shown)


DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+E.g. removing [`WorkerConfig`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConfig.java) would require removing [`OffsetBackingStore`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetBackingStore.java),
+which would require removing [`OffsetStorageReader`](https://github.com/apache/kafka/blob/3.6.0/connect/api/src/main/java/org/apache/kafka/connect/storage/OffsetStorageReader.java) etc., etc., resulting in substantial changes in the Debezium core and connectors.
+Therefore, this should be done in a separate task which would deserve a dedicated DDD describing all the changes and proposing replacements for Kafka Connect interfaces and classes.
+
+## Proposed changes
+
+### Thread pools
+
+Implementation of concurrent processing will be based on the Java [Executors](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html) framework.
+As the Debezium is currently based on Java 11, new concurrency features introduced by the project Loom, namely virtual threads and structured concurrency, cannot be used now.
```
jpechane, line 59: As we intend to move to Java 17 in next year, are there any Java 17 features that would help even more with the implementation and further improvements or it is not relevant for the new engine?

DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+Before passing a batch of events to the `ChangeConsumer`, user defined single message transformations are applied to the records.
+In the case of [`ConvertingEngineBuilder`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java#L53), events are moreover serialized to `SourceRecords` before passing to the `ChangeConsumer`.
+Both these tasks will be run in parallel in `ThreadPoolExecutor` mentioned above.
+Before passing the batch to the `ChangeConsumer`, tasks for all records in the batch will be awaited to finish.
+
+The implementation should also try to optimize record serialization/deserialization in case of the `ConvertingEngineBuilder`.
+Currently, the records are serialized to `SourceRecords` before passing to the `ChangeConsumer` and then deserialized back when `ChangeConsumer` calls `RecordCommitter`.
+TBD task is to find a way how to avoid record deserialization during commit without breaking existing API.
+
+#### Disabling total ordering of the CDC events
```
jpechane, line 102: There should be a note about offsets handling. We need to make sure with both parallel processing and even more with unordered delivery that only most earlier offsets are comitted. There should be two shutdown modes - one that will process remaining messages up to the latest delivered and another one immediate whose result will be increased number of duplicates upon connector restarts.

DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Run consumer function with engine task, e.g. in case of Kafka with {@link SourceTask}.
+     * Effectively expose engine internal task for testing.
+     */
+    void runWithTask(Consumer<SourceTask> consumer);
+}
+```
+
+In the future, we may add more methods to this interface if it would be beneficial from the testing point of view to expose any other `DebeziumEngine` internals or add convenient methods for testing.
+Number of such methods should be, however, kept as minimal as possible and methods should be added only for a very good reason, as the interface will force `DebeziumEngine` implementations to implement all these methods as well if the implementation should be tested with the Debezium testsuite.
```
jpechane, line 414: I'd like to see a reference to performance testing too.

jpechane (commented):
@vjuranek Good job!

DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+### Thread pools
+
+Implementation of concurrent processing will be based on the Java [Executors](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html) framework.
+As the Debezium is currently based on Java 11, new concurrency features introduced by the project Loom, namely virtual threads and structured concurrency, cannot be used now.
+However, it's expected to switch to virtual threads in the future, once Debezium will be based on Java 21 or higher.
+[`ThreadPoolExecutor`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html) will be used for the creating and pooling of the threads.
+The `ThreadPoolExecutor` will be created by factory method [`Executors.newFixedThreadPool​(int nThreads)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html#newFixedThreadPool(int)).
+Number of worker threads will be specified either by the user via configuration or the default value will be used.
+The default value will be the number of the cores of the underlying machine.
+There will be only one thread pool, the one for processing CDC event pipelines.
```
mfvitale: Is `processing CDC event pipelines` relates to `run potentially time-consuming code (e.g. event transformation or serialization) in the dedicated threads`?

DDD-7.md
vjuranek: I don't that there is anything relevant (at least not anything major).

DDD-7.md
vjuranek: yes, but CDC event pipeline I mean transform -> serialize -> (user handler) -> committer pipeline

DDD-7.md
vjuranek: Sure, completely forgot to mention it. The plan is to use JHM benchmark similar to `debezium-microbenchmark-oracle` with [`SimpleSourceConnector`](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/main/java/io/debezium/connector/simple/SimpleSourceConnector.java) or maybe add another implementation similar to this, more suitable for perf. testing (possibly with option to configure how large/complex generated records should be).

DDD-7.md
vjuranek: Good point - actually this is IMHO the hardest part of the whole implementation and it's not mentioned here as I haven't any good answer for it yet (I should probably add TBD section for it)

DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+* Remove dependency on Kafka Connect API.
+
+### Preserving Kafka Connect model
+
+As the main goal of the Debezium engine is to be able to execute Debezium outside of Kafka, it may be strange why not to take this opportunity to get rid of Kafka dependencies.
+The reason is simple: it would be a too complex change which would not impact only the Debezium engine.
+E.g. removing [`WorkerConfig`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConfig.java) would require removing [`OffsetBackingStore`](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetBackingStore.java),
+which would require removing [`OffsetStorageReader`](https://github.com/apache/kafka/blob/3.6.0/connect/api/src/main/java/org/apache/kafka/connect/storage/OffsetStorageReader.java) etc., etc., resulting in substantial changes in the Debezium core and connectors.
+Therefore, this should be done in a separate task which would deserve a dedicated DDD describing all the changes and proposing replacements for Kafka Connect interfaces and classes.
+
```
vjuranek, line 53: I've already started with implementation and one thing which complicates the code little bit is that Connect model doesn't allow access to connector and task contexts. Would it make sense to replace it with our interfaces and implementation as a preparation step to abandon Connect model?

It would look like this:
```
@Incubating
public interface DebeziumSourceConnectorContext {

    /**
     * Returns the {@link OffsetStorageReader} for this DebeziumConnectorContext.
     * @return the OffsetStorageReader for this connector.
     */
    OffsetStorageReader offsetStorageReader();

    /**
     * Returns the {@link OffsetStorageWriter} for this DebeziumConnectorContext.
     * @return the OffsetStorageWriter for this connector.
     */
    OffsetStorageWriter offsetStorageWriter();

    /**
     * Notifies the engine it should rise an error and allows engine to take appropriate action.
     * @param e Exception to be raised.
     */
    void onError(Exception e);
}
```
and implementation implementing both our and Connect interfaces:
```
public static final class EngineSourceConnectorContext implements DebeziumSourceConnectorContext, SourceConnectorContext
```
so that we have our own interface and at the same time we can still pass it to the other parts which require Connect model.

It's harder for part which are not interfaces, e.g. `SourceTask`, in such case we can have a delegate instance inside it:
```
public static final class EngineSourceConnector implements DebeziumSourceConnector {

        private final SourceConnector connectConnector;
        private DebeziumSourceConnectorContext context;

        public EngineSourceConnector(SourceConnector connectConnector) {
            this.connectConnector = connectConnector;
        }

        public SourceConnector connectConnector() {
            return connectConnector;
        }

        @Override
        public DebeziumSourceConnectorContext context() {
            return this.context;
        }

        @Override
        public void initialize(DebeziumSourceConnectorContext context) {
            this.context = context;
            this.connectConnector.initialize((SourceConnectorContext) context);
        }
    }
```
In a similar way we can probably handle storage part for now (as a preparation step - use our interface, implementation having also a Connect delegate and pass Connect delegate where needed).

Does it make sense to you @jpechane ? If yes, should I add it here or should be another DDD?

DDD-7.md
mfvitale: Maybe can be good to put it in the document.

DDD-7.md
```diff
@@ -0,0 +1,153 @@
...
+Providing a new implementation of `DebeziumEngine` interface which would run some of the tasks in parallel can give significant performance boost.
+New implementation and related changes should also aim for a good test coverage and ease of testing of the new implementation as well as any other future implementations.
+
+## Goals
+
+Provide new implementation of [`DebeziumEngine` interface](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java) which would
+
+* allow to run multiple [source tasks](https://github.com/apache/kafka/blob/3.6.0/connect/api/src/main/java/org/apache/kafka/connect/source/SourceTask.java) for given connector if the connector provides multiple tasks
+* run potentially time-consuming code (e.g. event transformation or serialization) in the dedicated threads
+* allow possible further speedup by optionally disabling total ordering of the messages
```
mimaison, line 27: These are interesting ideas that could potentially be implemented in Kafka Connect too. I created https://issues.apache.org/jira/browse/KAFKA-15912 to gauge the feasibility.

vjuranek: nice, thanks @mimaison !

DDD-7.md
vjuranek: will do @mfvitale

vjuranek:
In the past we had requests in the community if we are going to support multiple connectors in the Debezium server. Our answer was always "no", we want one connector per server. With future plan to have eventually some control plane over multiple servers. With this DDD I'm thinking about it again and I'm wondering if we should at least have open possibility to have multiple connectors under one engine/server. I'm thinking especially about the use case with one DB source connector and JDBC connector as the sink connector. IMHO this can be a good base for DB mirroring/migration/backup tool and no need to connect two independent processes would simplifies development, administration and there will be also performance gains. 

For now it would be mostly implementation detail, but when in it, why not to do it right now if we agree this is a good decision. As I mentioned [above](https://github.com/debezium/debezium-design-documents/pull/8#discussion_r1403175782), I'm creating Debezium connector and task abstractions and it would be natural to encapsulate most of the logic into these objects, e.g. in the engine `run()` method, instead of calling engine methods like
```
public void run() {
    ...
    startConnector(connector);
    createConnectorTasks(connector);
    startTasks(connector);
     ....
}
```

it would be something like this
```
public void run() {
    ...
    connector.start();
    connector.createTasks();
    connector.startTasks();
    ...
}
```

However, if we are sure we always want one connector per engine/server, it would be more natural to have most of the logic in the engine and `DebeziumSourceConnector` would act as a small support object (and eventually can be completely dropped).

Personally I'd prefer the first option (most of the logic in the connector and possibility to have multiple connectors in the future), with current implementation supporting only one connector.

WDYT?

jpechane:
@vjuranek TBH, I am really not fan of mutlple connectors support. Please keep in mind that the target of Debezium Server is to be smaller/simpler version of runtime, not Kafka Connect competition.
Your remark of the sink connector support is definitely valid but in that case we need to decide

1) Do we want to support JDBC sink connector? The intent already was that the JDBC sink will be slightly refactored and an API/SPI will be available to make it compatible with Debezium Server. Also if we'd like to intorduce a full parallel processing from transaction log to sink write then we might end up in skipping Kafka Connect API on the source side completely and we'll interact with Debezium framework directly. Also it'd be a bit strange to support JDBC sinks as we do now and then support JDBC sink in a different way.
2) If we want to support (or not support but provide an ability for) for arbitrary sink connector then maybe we should intorduce a Debezium Server sink adapter that will fit into sink implementation and serve as translator/integrator between Debezium Server sinka and Kafka Connect sink API. That way we could intorduce early support for JDBC sink and later on implement option 1).

vjuranek:
Thanks @jpechane for the comments. I didn't mentioned it in my previous comment, but yes, the main drawback I see with this approach is that sooner or later we will end up with re-implementing most of the Kafka stuff, so I completely agree  on this part.
As for JDBC sink support - in general (read: in undefined future) I'd say yes, IMHO we should support it (the logic behind it is that we should provide the same level of support as for other connectors and as we provide runtime for source connectors, we should do it also for the sink). But definitely would be good to wait for the community first to see if there is any interest for it at all.

Anyway, I have no strong opinion about that, so no problem with keeping most of the logic in the engine itself. If we later on decide to support also sink connector, it should be quite easy do the refactoring.

DDD-7.md
```diff
@@ -0,0 +1,421 @@
...
+Reading the offsets from multiple tasks therefore shouldn't introduce any concurrency issues and for writing [OffsetStorageWriter](https://github.com/apache/kafka/blob/3.6.0/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetStorageWriter.java#L65) is explicitly marked as thread-safe.
+
+### Processing CDC events concurrently
+
+Current `DebeziumEngine` API partially delegates event processing to the [`ChangeConsumer`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java#L159) implementation, which is provided by the user.
+More specifically, events are passed to the [`ChangeConsumer.handleBatch()`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java#L167) in a batch manner, as a `List`.
+This makes sense as in many cases events are submitted to another system and this is usually more efficient to do in a batch.
+However, this prevents us from making a complete event processing pipeline, which would be run in a dedicated thread.
```
aleksandervalle, line 87: Since you mentioned community feedback, here's one :smile: Apologies if this is outside scope of this design document.

I assume that a large chunk of the time an event uses from source to sink is spent travelling from Debezium to the sink (and the response back to Debezium), especially if the sink is physically located far away from where Debezium is running.

For sinks that maintain ordering of events within partitions, like Kafka and Azure Event Hubs, it seems like one could gain performance if the sink adapter grouped a batch by destination topic, then by partition id, and sent those groups in parallel to the sink? I guess is would depend on batch size and latency between Debezium and sink.

vjuranek: hi @aleksandervalle , thanks for the comment. Yes, I totally agree, however, this is not anything we can do in the engine itself as it doesn't know how the sink will look like. This needs to be done in Debezium server, namely in concrete sink implementations.

jeremy-l-ford:
Is there a possibility that the engine could provide both parallelism and ordering?  In a manner akin to how Kafka handles its partitions, each message would get hashed by its key.  The hashes would be assigned to a queue/thread.  For the use case of an event whose key is the table name, this would gain parallelism while maintaining order at the table level.

Perhaps a pluggable impl to determine the hashes/queue id, where 

- serial = 1 queue defined and one hash / queue Id always returned
- no order =  N number of queues and random or smallest queue hash/id returned
- hash based = N number of queues where message key determines hash / queue id

Perhaps this makes things overly complex though.  Anyways, just a thought.

vjuranek:
Thanks for your input @jeremy-l-ford , I think this a very good idea! I'll think more about it, but it shouldn't be that hard to implement it.

We probably cannot use it as the default and should still provide total order of messages as the default (e.g. when one table references values in another table with foreign key constraint, insert in such table would fail if it arrives sooner than the insert in referenced table), but it can be another option between running consumer in a single thread and running consumer in multiple threads. Basically what we would need here is to expose `ExecutorService` (or, in this case, set of `ExecutorService`s created as `newSingleThreadExecutor()`) to `RecordProcessor`, which would be little bit dangerous as we would pass `ExecutorService` to potentially unknown code (which may either shut down the service or handle exception in an unexpected way), but the benefits may outweigh the risks. Or possibly we can expose `ExcutorService` only to our implementation of `RecordProcessor` and if we decide in the future to provide `RecordProcessor` via SPI, we won't provide `ExecutorService` to these implementations.

Any thoughts on this @jpechane ?

DDD-7.md
```diff
@@ -0,0 +1,421 @@
...
+     * @return the DebeziumSourceTaskContext for this task
+     */
+    DebeziumSourceTaskContext context();
+}
+```
+
+```java
+@Incubating
+public interface DebeziumSourceTaskContext {
+    /**
```
vjuranek, line 330: This needs to be consolidated with [`CdcSourceTaskContex`](https://github.com/debezium/debezium/blob/v2.4.0.Final/debezium-core/src/main/java/io/debezium/connector/common/CdcSourceTaskContext.java)

DDD-7.md
```diff
@@ -0,0 +1,421 @@
...
+For more detail about possible issues, please see [DBZ-2534](https://issues.redhat.com/browse/DBZ-2534).
+To prevent this situation, transition from `STARING_TASKS` into  `STOPPING`  won't be possible by calling engine's `close()` method.
+Also, `STARING_TASKS` must finish completely.
+Even if one of the threads fails to start the task it was running, the main (engine) thread has to wait for all other tasks until they finish (no matter if successfully or not) before moving into the `STOPPING` state.
+In general, transition `STARING_TASKS` ->  `STOPPING` is possible, but it should happen only in the case when an exception was thrown from method starting tasks and only until all threads starting tasks have finished.
+
+#### Exception handling
+
+Retriable exceptions are handled at the place where they happen and relevant action is retried until `ERRORS_MAX_RETRIES` attempts are exhausted.
+Contrary to the existing `EmebeddedEngine` implementation, task is not restarted at this point (**TODO**: *re-think, why is the task restarted in the current implementation?*).
```
vjuranek, line 233: Restart loop was added into embedded engine in [11c588fb0](https://github.com/debezium/debezium/commit/11c588fb0350495fe8eb5a085ac9791b1b1f4c53#diff-e814dc964227048cece0a87ca5a8290f86c6d3a07cf7080c3dc880006dad1ab9R841-R866). Justification in related [DBZ-4629](https://issues.redhat.com/browse/DBZ-4629) is not very clear, but as per [this comment](https://github.com/debezium/debezium/pull/3857/files#r990926825), it seems that the main reason was that some connectors thrown exceptions from their `start()` method which weren't retriable. We have most of the exception retriable now (since [DBZ-5244](https://issues.redhat.com/browse/DBZ-5244)), but [this comment](https://github.com/debezium/debezium/pull/3857#issuecomment-1245313387) claims it doesn't solve the issue. It's not clear to me which issue exactly it doesn't solve, but as we stop and restart the task in `poll()` method of `BaseSourceTask`, I don't believe we need to do it in engine again (currently, retries are applied only for exception thrown from `task.poll()` call). I believe we just need to catch `RetriableException` thrown from `BaseSourceTask::poll()` method and let the engine to call `BaseSourceTask::poll()` again, which should result into calling `BaseSourceTask::startIfNeededAndPossible()` which should eventually start the task again. If there is any issue with it, it should be IMHO solved in the right place where the issue actually is and adding additional task stop/start in embedded engine shouldn't be needed.

DDD-7.md
jpechane: Let's add it here

PR #10, opened 2024-05-29
DBZ-7234: Design document for read only incremental snapshot

mfvitale:
The aim of this DDD is to explore a possible way to have read only incremental snapshot for others databases apart from MySQL.

In this document there is a proposal for PostgreSQL. Others will follow.


DDD-8.md
```diff
@@ -0,0 +1,118 @@
+# DDD-7: Read-only incremental snapshots for other relational connectors
```
vjuranek: s/DDD-7/DDD-8/

DDD-8.md
```diff
@@ -0,0 +1,118 @@
...
+The figure below shows the process of deduplication. We have:
+
+* Transaction log: is the resulting database logs, each square represent and event with the indication of: its key, the type of operation (**I**nsert, **U**pdate, **D**elete), and the associated transaction.
+* Back-fill chunks: record read from the database during chunk snapshot. Striped one means that the record has been removed by the deduplication.
+* Output stream: the final emitted events. 
+
+The think to note here is what happens to the record with key `k2`. 
+This record is read by the snapshot and, since it associated to a transaction that completed during the snapshot - the high watermark reveal only the `Tx13` is still in progress, 
+we are sure that this event must be deduplicated. This will effectively remove from the windows the event that was read.
+
```
vjuranek, line 100: I'd add short info what happens when TX starts after select starts and finishes before select finishes. `xid` should be always growing, so IMHO we should be fine also in this case.
May also worth to mention that documentation says that the function uses `xid8` which never wraps around so we don't need to take case about it either.

mfvitale: > I'd add short info what happens when TX starts after select starts and finishes before select finishes. xid should be always growing, so IMHO we should be fine also in this case.
May also worth to mention that documentation says that the function uses xid8 which never wraps around so we don't need to take case about it either.

Good points. I'll add something about them.

jpechane (approved):
@mfvitale Nice job, thanks, this looks really feasible!

DDD-8.md
mfvitale: > I'd add short info what happens when TX starts after select starts and finishes before select finishes. xid should be always growing, so IMHO we should be fine also in this case.

If I understood you are talking about the following scenario:

PostgreSQL read only (7)


And if so, I think the current proposal should not work. The `tx12` was not running while getting the list of running transactions and so when it arrive to the processing, even if the window is open, it will not be de-duplicated due to

```text
if completedTxs.contains(e.txId)
```
Since the `tx12` started and finished during the snapshot and was not detected as a in-progress transaction.

I think then the algorithm should be

```text
(1) Set lwInProgressTx := xip from pg_current_snapshot()
(2) chunk := select next chunk from table
(3) Set hwInProgressTx := xip from pg_current_snapshot()
  inwindow := false
  // other steps of event processing loop
  while true do
       e := next event from changelog
       append e to outputbuffer
       if not inwindow then
           if lwInProgressTxSet.contains(e.txId) //reached the low watermark
               inwindow := true
       
       if not hwInProgressTx.contains(e.txId) //haven't reached the high watermark yet
           if chunk contains e.key then
               remove e.key from chunk
       else //reached the high watermark
           for each row in chunk do
               append row to outputbuffer
   // other steps of event processing loop
```
@vjuranek  WDYT?

vjuranek: @mfvitale now you don't use `lwInProgressTx` for deduplication. IMHO the original algorithm was right, you just need to specify what `hwInProgressTxSet subtracted by lwInProgressTx` does. If it's substraction of the sets, it's wrong as it would miss `Tx12`, but if it's substraction in a sense that the result would be a set with lower bound of  `pg_snaphot_min(lwInProgressTx)` and higher bound of `max(pg_snaphot_max(lwInProgressTx), pg_snaphot_max(hwInProgressTxSet)`, it's IMHO correct.

Thinking more about it, I wonder if we need to take into account also current connection isolation level. If I'm not mistaken, incremental snapshot don't set any specific isolation level (I didn't check, if it does we are fine) so we may also need to deduplicate records from `Tx14` from the picture, as it hasn't been committed yet, but with isolation level `READ UNCOMMITTED` we still may see changes from this Tx (e.g. returning new record create by Tx and later on get create event for this record when Tx is committed). IMHO the above approach should cover also this case (`max(pg_snaphot_max(lwInProgressTx), pg_snaphot_max(hwInProgressTxSet)` should return 14, so we would deduplicate also records with keys from Tx 14), but it would IMHO deserve few sentences of explanation in the design doc (and also double check if I'm right or not:-). Another approach could be to set explicitly isolation level for connection reading the data during incremental snapshot.

mfvitale: > now you don't use lwInProgressTx for deduplication. 

Yes and this solve the issue with `Tx12` since the original idea was to just subtract the sets, but in this way transaction that was not in-progress during the low watermark, but started after, will be not considered. 

Looking to the last algorithm it has an issue on closing the windows. If all transactions that are  in-progress during the high watermarking query will never be committed, then the algorithm will never close the windows and will continue to de-duplicate.  

So effectively I was arrived to your conclusion: relaying on the min and max will be the better option since the txId is an increasing monotone function we have:

* txid >  `pg_snaphot_min(lwInProgressTx)`  -> inside the snapshot window, de-duplication needed
* txid >  `max(pg_snaphot_max(lwInProgressTx), pg_snaphot_max(hwInProgressTxSet))`  -> out of the window, stop de-deduplication

> If I'm not mistaken, incremental snapshot don't set any specific isolation level (I didn't check, if it does we are fine) 

Yes, we don't explicitly set it during incremental snapshot. 

> so we may also need to deduplicate records from Tx14 from the picture, as it hasn't been committed yet, but with isolation level READ UNCOMMITTED we still may see changes from this Tx (e.g. returning new record create by Tx and later on get create event for this record when Tx is committed).

Seems to me that the default is [READ COMMITED](https://www.postgresql.org/docs/current/transaction-iso.html) so in that case we should be fine. 

So the final algorithm should be 

```text
(1) pg_snapshot lwSnapshot := pg_current_snapshot()
(2) chunk := select next chunk from table
(3) pg_snapshot hwSnapshot := pg_current_snapshot()
  inwindow := false
  // other steps of event processing loop
  while true do
       e := next event from changelog
       append e to outputbuffer
       if not inwindow then
           if e.txId >= pg_snaphot_min(lwSnapshot)  //reached the low watermark
               inwindow := true
       
       if e.txId < max(pg_snaphot_max(lwSnapshot), pg_snaphot_max(hwSnapshot)) //haven't reached the high watermark yet
           if chunk contains e.key then
               remove e.key from chunk
       else //reached the high watermark
           for each row in chunk do
               append row to outputbuffer
   // other steps of event processing loop
```

vjuranek: > > now you don't use lwInProgressTx for deduplication.
> 
> Yes and this solve the issue with `Tx12` since the original idea was to just subtract the sets, but in this way transaction that was not in-progress during the low watermark, but started after, will be not considered.

Agree, my point here was that we may try to de-duplicate already committed transactions, see bellow.

> > so we may also need to deduplicate records from Tx14 from the picture, as it hasn't been committed yet, but with isolation level READ UNCOMMITTED we still may see changes from this Tx (e.g. returning new record create by Tx and later on get create event for this record when Tx is committed).
> 
> Seems to me that the default is [READ COMMITED](https://www.postgresql.org/docs/current/transaction-iso.html) so in that case we should be fine.

Yes, but the use may change it. IMHO it would be quite rare, so I think this can be skipped for initial implementation and implemented only when users want it, but I'd mention is as a known limitation in the design doc as well as in the documentation.
 
> So the final algorithm should be
> 
> ```
> (1) pg_snapshot lwSnapshot := pg_current_snapshot()
> (2) chunk := select next chunk from table
> (3) pg_snapshot hwSnapshot := pg_current_snapshot()
>   inwindow := false
>   // other steps of event processing loop
>   while true do
>        e := next event from changelog
>        append e to outputbuffer
>        if not inwindow then
>            if e.txId >= pg_snaphot_min(lwSnapshot)  //reached the low watermark
>                inwindow := true
>        
>        if e.txId < max(pg_snaphot_max(lwSnapshot), pg_snaphot_max(hwSnapshot)) //haven't reached the high watermark yet

IMHO the condition should be `e.txId < max(pg_snaphot_max(lwSnapshot), pg_snaphot_max(hwSnapshot)) && e.txId > pg_snaphot_min(lwSnapshot)` otherwise we may de-duplicate already committed transactions (e.g. Tx10 on the image) when the record is processed after we read the table from the snapshot connection.

>            if chunk contains e.key then
>                remove e.key from chunk
>        else //reached the high watermark
>            for each row in chunk do
>                append row to outputbuffer
>    // other steps of event processing loop
> ```

mfvitale: > Yes, but the use may change it. IMHO it would be quite rare, so I think this can be skipped for initial implementation and implemented only when users want it, but I'd mention is as a known limitation in the design doc as well as in the documentation.

Correct, I was talking about the current implementation but you are right, I'll add this as a limitation point. 

> IMHO the condition should be e.txId < max(pg_snaphot_max(lwSnapshot), pg_snaphot_max(hwSnapshot)) && e.txId > pg_snaphot_min(lwSnapshot) otherwise we may de-duplicate already committed transactions

Not sure I got your point. The `Tx10` transaction should open the de-duplication window and, after that, any further transaction will be de-duplicated until the high watermark will be reached. Am I missing something?

vjuranek: > Not sure I got your point. The `Tx10` transaction should open the de-duplication window and, after that, any further transaction will be de-duplicated until the high watermark will be reached. Am I missing something?

Sorry, my bad, there cannot be any events from TX10 because it's already committed. So yes, you are right and the condition should be as you had it: `if e.txId < max(pg_snaphot_max(lwSnapshot), pg_snaphot_max(hwSnapshot))`

vjuranek (approved):
LGTM now, nice work @mfvitale

PR #11, opened 2024-07-12
DBZ-7234 DDD-8 Add MongoDB

mfvitale:


DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+Since Standalone server do not have the _oplog_ and so Debezium is not able to streaming events from it, we do not consider it. Although MongoDB does not recommend running a standalone server in production.
+
+##### Replica set
+
+A [replica set (RS)](https://www.mongodb.com/docs/manual/core/replica-set-members/) in MongoDB is a group of _mongod_ processes that provide redundancy and high availability. The members of a replica set are:
+
+* Primary:
+  receives all write operations.
+* Secondaries:
+  replicate operations from the primary to maintain an identical data set. To replicate data, a secondary applies operations from the primary's oplog to its own data set in an asynchronous process. A replica set can have one or more secondaries.
```
jcechace, line 151: Technically there are also hidden members (secondaries which are purposefully hidden from client advertising) and arbiter nodes (nodes which don't hold data but vote in primary election). Neither of these are likely relevant for our case (although technically a hidden node could be used for CDC with direct connection)

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+
+By default, an application directs its [read operations](https://www.mongodb.com/docs/manual/core/read-preference/) to the primary member in a replica set (i.e. read preference mode "primary"). But, clients can specify a read preference to send read operations to secondaries.
+
+This type of configuration is common for CDC scenario where users wants to reduce the load on the primary that is already under the write load.
+
+So you can directly connect to a secondary specifying its address and setting the read preference mode accordingly. 
+
+This configuration can lead to a possible weird issues in particular type of deployments. For example, if you have that nodes of the RS has two network address, internal and external, and you configure 
+the connection string to point to a secondary using its external address, when the client connects to the RS and retrieve its information, even the intern addresses of the nodes, it then can use the internal address causing connection failures. 
+To overcome to this, you can set the [directConnection](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.directConnection) property.
```
jcechace: More specifically the client will always use the addresses returned by the cluster description. For this reason with RS topology it is generally not a good idea to configure the RS with addresses which are not reachable form the outside. The correct approach in this case is to use hostnames and make sure that the hostname is resolved to internal address for in-cluster communication and resolve to a public address for external access.

Note that `directConenction` requires only a single host and sacrifices HA capabilities (if the node goes down the client will not switch to a different node in the RS).

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+
+The [oplog](https://www.mongodb.com/docs/manual/core/replica-set-oplog/) (operations log) is a special capped collection that keeps a rolling record of all operations that modify the data stored in your databases.
+If write operations do not modify any data or fail, they do not create oplog entries. 
+
+MongoDB applies database operations on the primary and then records the operations on the primary's oplog. The secondary members then copy and apply these operations in an asynchronous process. All replica set members contain a copy of the oplog, in the local.oplog.rs collection, which allows them to maintain the current state of the database.
+To facilitate replication, all replica set members send heartbeats (pings) to all other members. Any secondary member can import oplog entries from any other member.
+
+Change streams allow applications to access real-time data changes without the prior complexity and risk of manually tailing the oplog. Applications can use change streams to subscribe to all data changes on a single collection, a database, or an entire deployment, and immediately react to them.
+Change streams is used by MongoDB Debezium connector.
```
jcechace, line 206: Note that mongos itself has no oplog, however it still can be used to open a change stream. Under the hood mongos will open change stream against each shard, synchronise the events and serve them to the client via the change stream opened against mongos.

The need for synchronisation can be a source of performance degradation if there is a so called "cold shard" (a shard with minimal write) as mongos need to check whether the shard received any new changes -- no change is noted on the oplog via periodic noop entry in the oplog. So potentially the synchronisation can be delayed until the next noop is written. See https://www.mongodb.com/docs/manual/administration/change-streams-production-recommendations/#sharded-clusters

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+For example, in a scenario where the Debezium user want to connect directly to a replica, we can just read the _oplog_ of that specific node that can have inconsistent information due to the asynchronous nature of the data replication between primary and secondary nodes. 
+
+This in a RS set deployment, and it will be completely not possible in a sharded cluster deployment, since the connection goes through the _mongos_ component.
+
+To solve this limitation we can think to directly use the change stream API instead going directly on the _oplog_. 
+So we will have a change stream for the normal event streaming and a temporary change stream opened specifically to get the low and high watermarks.
+
+```java
+MongoCursor<ChangeStreamDocument<Document>> cursor = db.watch().iterator();
+ChangeStreamDocument<Document> lastDocument = cursor.next();
```
jcechace: `next()` will block until the next event arrives -- which could be a while -- and fails on timeout. We should use `tryNext()`

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+
+BsonTimestamp tLast = lastDocument.getClusterTime();
```
jcechace: Personally I wouldn't use the cluster time. I didn't find any specific documentation, however I suspect that there might be a difference between cluster time and the time actually encoded in the resume token. For this reason I would extract the timestamp from the resume token's `_data` field instead

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+Furthermore, based on the different type of MongoDB topologies there can be problems to have a consistent information reading directly the _oplog_. 
+For example, in a scenario where the Debezium user want to connect directly to a replica, we can just read the _oplog_ of that specific node that can have inconsistent information due to the asynchronous nature of the data replication between primary and secondary nodes. 
+
+This in a RS set deployment, and it will be completely not possible in a sharded cluster deployment, since the connection goes through the _mongos_ component.
+
+To solve this limitation we can think to directly use the change stream API instead going directly on the _oplog_. 
+So we will have a change stream for the normal event streaming and a temporary change stream opened specifically to get the low and high watermarks.
+
+```java
+MongoCursor<ChangeStreamDocument<Document>> cursor = db.watch().iterator();
```
jcechace: IMHO calling `cursor()` instead of `iterator()` is more idiomatic. Technically it doesn't matter as `iterator()` on `ChangeStreamIterableImpl` just calls `cursor()`

DDD-8.md
```diff
@@ -126,3 +126,146 @@ So if the snapshot is reading a chunk from a table that is modified through a su
...
+
+BsonTimestamp tLast = lastDocument.getClusterTime();
+```
+
+In case no document is present we can execute a fake operation to get the operation timestamp.
+
+```java
+var database = client.getDatabase(dbName);
+var result = database.runCommand(new Document("hello", 1), BsonDocument.class);
+BsonTimestamp tLast = result.getTimestamp("operationTime");
```
jcechace: This is potentially problematic as there might be operations in the meantime. I'd suggest a different approach. 

Once change stream  cursor is iterated on -- that is when `next()` or `tryNext` was called, the cursor is guaranteed to return a resume token when `cursor.getResumeToken()` is called (see `MongoChangeStreamCursor` interface). Thus when `tryNext()` returns a document we get the resume token by calling `doc.getResumeToken` otherwise we call `cursor.getResumeToken()` (this is exactly what the our streaming does currently.

jcechace (commented):
Left few clarifications. Overall a good approach IMHO. Just to few modifications

1) I would store the entire resume tokens as the low/high bounds
2) When Snapshot signal is received the streaming MUST be paused (as we need to ensure that streaming will not get ahead of the fresh change stream used to get the lower bound). 
3) Then a resume token is retrevied from a fresh change stream (either as the token of the next change stream document or the post-batch resume token of the cursor). This is stored as the lower bound
4) Streaming can be safely resumed. 


Now when it comes to detecting whether we should start buffering



1) First simply compare the resume token of current streaming iteration with the lower bound resume token. If it matches, we have reached the lower bound.
2) If it doesn't match we extract the timestamp from both tokens and check whether current streaming timestamp is the same or later than the lower bound's timestamp.

mfvitale:
> 1. First simply compare the resume token of current streaming iteration with the lower bound resume token. If it matches, we have reached the lower bound.

This will not work if the event with the exact resume token is filtered. I prefer to just compere if the event has a greater timestamp.

jcechace:
> > 1. First simply compare the resume token of current streaming iteration with the lower bound resume token. If it matches, we have reached the lower bound.
> 
> This will not work if the event with the exact resume token is filtered. I prefer to just compere if the event has a greater timestamp.

As discussed on Zulip -- this is not true as filtering for MongoDb is done on the DB side. Nevertheless, comparing the timestamps is sufficient as the original proposed comparison is logically equal to just comparing timestamps. 

Otherwise LGTM, thanks for putting this together!

mfvitale:
@jpechane Can you please have a look?

PR #12, opened 2024-08-08
DDD-4 Manipulate Connector Offset via Debezium Signals

ani-sha:
This is the first draft of offset manipulation design document (simplified version of #6), starting with implementation for Debezium Postgres connector.

cc : @jpechane

### DIFF
--- a/DDD-4.md
+++ b/DDD-4.md
@@ -1,0 +1,168 @@
+# DDD-4: Manipulate Connector Offset via Debezium Signals
+
+## Motivation
+
+Debezium is a platform to stream changes executed on data by processing transaction logs of a database.
+As changes are processed, a position into the transaction log is maintained so that Debezium can safely resume on restarts or failures.
+
+This position, often referred to as offsets, differs by connector. While the MySQL connector maintains a reference to the binlog filename and a position within the binlog, other connectors maintain state that is specific to that database vendor's implementation.
+Therefore, there is no single common structure for these offsets.
+
+Sometimes it is necessary to manipulate these offsets, for example, to reset a connector to a specific point in time or to resume from a specific point in time. There has been a long-standing need to provide a common way of managing these offsets, and this document proposes a solution to this problem. Until now the user had to manage the offsets themselves, which is error-prone and cumbersome.
+
+This document proposes a solution to this problem by using Debezium signals to handle the offset management in a common way. This will allow users to manage the offsets in a consistent way across all connectors.
+
+## Goals
+
+* Provide support reading connector offsets
+* Provide support writing connector offsets
+
+
+## Current connector offset behavior
+
+Debezium 3.x supports the notion of partition-based offsets based on the following contracts:
+
+* `io.debezium.pipeline.spi.OffsetContext`
+* `io.debezium.pipeline.spi.OffsetContext$Loader`
+* `io.debezium.pipeline.spi.Partition`
+* `io.debezium.pipeline.spi.Partition$Provider`
+* `io.debezium.connector.common.OffsetReader`
+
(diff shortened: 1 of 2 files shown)


DDD-4.md
```diff
@@ -0,0 +1,168 @@
...
+  "incremental_snapshot_collections": "<json>" // string
+}
+```
+The incremental snapshot context will serialize the collections that are still to be incrementally snapshot to the connector offsets as JSON and store this serialized value in the `incremental_snapshot_collections` field.
+This field is a JSON array of entries that have the following format:
+
+```json
+{
+  "incremental_snapshot_collections_id": "<database>.<schema>.<table>", // string, the TableId
+  "incremental_snapshot_collections_additional_condication": "ID = 25", // string, the additional where condition, may be null
```
mfvitale:
```suggestion
  "incremental_snapshot_collections_additional_condition": "ID = 25", // string, the additional where condition, may be null
```

DDD-4.md
```diff
@@ -0,0 +1,168 @@
...
+
+Provide a signal action to read and write offsets for each connector. The signal action will be a new action type in the Debezium signal framework.
+
+### Signal Action
+
+The `Change Offset` signal action will be used to change the offset of a connector. The action will have the following attributes:
+
+* `name` - The name of the action. This will be `change-offset`.
+* `offset-position` - The new offset position to set (which corresponds to `lsn_commit` in the PostgreSQL connector).
+* `last-offset-position` - The last offset position to set (which corresponds to `lsn_proc` in the PostgreSQL connector).
```
jpechane: I think we need to come up with better naming and descrption/explanation, current one is a bit confusing.
Also I think we need an example of the signal message in the DDD too.

DDD-4.md
```diff
@@ -0,0 +1,168 @@
...
+
+The PostgreSQL connector maintains offsets that align with our relational connectors.
+The offsets consist of a set of state that differs based on whether the connector is in a snapshot or streaming phase.
+
+This connector maintains a single partition of offsets, based on `database.dbname`.
+
+The following refers to the offsets regardless of streaming/snapshot:
+
+```json
+{
```
jpechane, line 47: We should also make a distintion between the fields that are an the offsets and those that are actually necessary for repositioning.

DDD-4.md
Naros: I don't think we should tie the `change-offset` payload to any specific connector implementation.

While there could be some common attributes between them, I believe it's best to take this as an iterative approach where we consider the payload being processed by the signal handler and delegated to each connector with a specific impl that validates & returns the writable payload.  Then we can look to refactor that as a second pass, if needed.

So I would propose the signal payload may look:
```json
{
  /* This acts as an envelope where we can add some common attributes in pass 2 */
  "connector-offsets": {
    /* This is the connector-offset payload, to be processed by each connector individually */
    /* Here we would have unique key/value tuples, this being an example of PostgreSQL */
    "last_commit_lsn": "....",
    "last_procsesed_lsn": "...",
    
    /* Here's another example for something like Oracle */
    "last_processed_scn": "123456789",
    "committed_scns": [
      {
        "redo_thread": 1,
        "commit_scn": "123456799",
        "transactions_committed_at_commit_scn": ["trx1", "trx2"]
      },
      {
        "redo_thread": 2,
        "commit_scn": "123456650",
        "transactions_committed_at_commit_scn": ["trx3"]
      }
    ]
  }
}
```

The envelope could be something handled directly by the main signal handler if needed, but we'd have a connector-specific object that is provided the contents of the `"connector-offsets"` and it would return the necessary Struct payload based on that data to be written to the Kafka offsets.

In this case, you may not need the `OffsetValidator` contract to be separate, too.  wdyt?

PR #14, opened 2025-03-01
DDD-9 A buffer-less implementation of Oracle LogMiner

Naros:


DDD-9.md
```diff
@@ -0,0 +1,111 @@
...
+Currently, the Oracle connector maintains two critical values in the offsets, `scn` and `commit_scn`, shown here:
+
+```json
+{
+    "scn": "123456",
+    "commit_scn": "123456:1:txId1-txId2,123457:2:txId4"
+}
+```
+
+Moving forward, the connector needs to maintain the resume SCN and each committed transaction id since the resume SCN and its commit SCN.
```
jpechane: How large this data can be? W need to make ssure will not overload the offset handling mechanism.

DDD-9.md
```diff
@@ -0,0 +1,111 @@
...
+In addition, need to maintain the last processed transaction id and event sequence.
+I would also prefer to move away from _encoded_ values to the more natural nested `Struct`s style.
+
+I'm not sure how hard this makes editing the offsets for users, but this what I have in mind:
```
jpechane: Offset is just a `Map`. This means you'd need to generate the JSON/structure yourself and put it under a single key.
I don't think that map of maps is supported.

DDD-9.md
```diff
@@ -0,0 +1,111 @@
...
+### LogMiner query
+
+The LogMiner query must be adjusted to include a new predicate condition to exclude transactions in `committed_transactions`.
+The query change would include something like the following:
+
+```sql
+AND UPPER(RAWTOHEX(XID)) NOT IN ( <committed-transaction.id(s) in upper-case> )
+```
+
+The construction of the in-clause should use the same logic for the `table.include.list` to wrap the predicate in a disjunction is the list of values exceeds 1000 values, since Oracle does not permit creating in-clauses with more than 1000 elements.
```
jpechane:
```suggestion
The construction of the in-clause should use the same logic for the `table.include.list` to wrap the predicate in a disjunction if the list of values exceeds 1000 values, since Oracle does not permit creating in-clauses with more than 1000 elements.
```

DDD-9.md
```diff
@@ -0,0 +1,111 @@
...
+TBD
+
+## Concerns / Gaps
+
+* LogMiner will reorder all changes in chronological commit order, meaning SCN values will increment/decrement from row-to-row, while the commits will always be in chronological order, always equal-to or greater.
+* DDL changes need to be reviewed to make sure they do not introduce any unexpected/unintended consequences of dispatch/processing order.
+* Large transactions will be re-read across mining steps; will this create any disk IO performance concerns on the database?
+* Will this be a problem for any of the three current mining strategies using committed data?
+* What about partial rollbacks, do we still need to handle these?
+* Do we implement this as a new adapter, i.e. `logminer_slim`, or change `logminer` proper?
```
jpechane: I'd propose a new adapter.

DDD-9.md
```diff
@@ -0,0 +1,111 @@
...
+* Is the change in offsets being non-backward compatible a blocker or concern?
```
jpechane: There should be a migration path from old offsets to the new adapter when old adapter is kept as default.
W'll need to discuss the need for new -> back migration path only when the new adapter will become new default.

DDD-9.md
Naros: It's tough to say as it would depend on the concurrency and duration of overlapping transactions. 

In hindsight, we may be able to rely on the `COMMIT_SCN` field in the view if LogMiner populates the column reliably and consistently. This would enable us to continue to track thread-specific commit details as we do now, and limit the number of transaction ids stored to only those that share the same commit scn.

In this case we could perhaps build a fairly reasonable SQL predicate that for a given redo thread, we exclude any transactions that commit prior to the redo thread's offset `commit_scn` value, and any transactions within the tracked transaction id list with the `commit_scn` value.

That would be worth a shot, and then I think `commit_scn` values in the offset could be retained as-is, and we could also keep the `scn` the same, just how its value is resolved in the connector differs.

Then its really only the tracking of the current dispatched transaction & its event sequence :+1: that comes down as an offset difference (if we want to keep the offset format identical for compatibility).

DDD-9.md
Naros: In that case, whether we use an encoded string or JSON is just a mere implementation detail and boils down to what is easier for users to supply in the end.

DDD-9.md
Naros: Based on the above conversations, if map of map isn't supported and the internal `commit_scn` representation is an implementation detail, we could probably avoid this being an issue altogether for both directions.

PR #15, opened 2025-04-04
DDD-13 DBZ#1204 Debezium component descriptors

### LINKED ISSUE #1204: Debezium component descriptors / registry [DBZ-8420]
Migrated from [DBZ-8420](https://issues.redhat.com/browse/DBZ-8420)

The idea is to propose and create a set of descriptors which would describe the configuration and components available for a particular distribution of debezium. This can be thought of as extension of current configuration schemas we produce for source connectors -- however the format should be revisited as OAPI seems a bit too heavy.

In regards to DS the idea is to have "registry" which would describe all components and their configuration available for particular distribution of DS. As a tree strucutre this could look like follows

```text
index.json 
|_sources
   |_mongo.json
   |_mysql.json
|_sinks
   |_ redis.json
|_ transforms
   |_outbox.json
|_storages?
```

The entire tree should be available via http url. The entry point should likely be the index.json file which would contain the required information about the distribution (metadata such as version, images,... as well as link to descriptors for individual components). Technically all information could be in a single file, however the goal is to also have this human readable. 

In the end we should be able c produce these in automated manner, however such descriptors also should be easy to write manually as user's might want to add e.g. custom transformations. 

Platform backend will be able to serve these to frontend which will use them to render UI.

mfvitale:
Initial draft to start the discussion on https://issues.redhat.com/browse/DBZ-8420


Naros:
Hi @mfvitale, I'm not opposed to using the `schema-generator`; however, I do want us to reevaluate a few changes to how this generator gets the data it needs.

**Field Attributes**

This is probably the most brittle aspect of the Debezium `Field` meta-model. 

Anytime a new configuration is added, and if that configuration makes sense to be rendered next to other existing configurations in that group in the UI, this mandates a significant change in tracking down all the configurations that need their group order adjusted. This is not only time-consuming, it's error-prone. 

Field width & importance are also things that are very UI-centric. I again understand why we added these to the code initially, but I do wonder if these along with the group order should exist elsewhere.

**Localization/I18n**

It's inevitable, we will be asked about Internationalization (i18n) support for other localizations. I mentioned this to Indra sometime ago, and we should at least think about how we might consider supporting it. One idea here would be to use the `displayName` and `description` values from the `Field` descriptor as the fallback, but have an alternative that would drive locale-specific translations for fields, descriptions, etc. The `Field` descriptor in the code base is certainly not the most ideal solution for this when there are far better tooling for this imo.

I'm not yet sure what would be the best way to do this outside of the Java code, but I'd like to at least proposal we think about this and see if there is a reasonable, low-cost way to accomplish this for future maintainability.

mfvitale:
@Naros thanks for your feedback. 

> Anytime a new configuration is added, and if that configuration makes sense to be rendered next to other existing configurations in that group in the UI, this mandates a significant change in tracking down all the configurations that need their group order adjusted. This is not only time-consuming, it's error-prone.

The problem is not just ours, the `ConfigKey` from Kafka Connect has the same concept: 'group` and 'orderInGroup`. 
I agree that this concept are more UI related stuff but whenever you put these you will have to "order" in same way. 

You can think for example to put just the UI related stuff separately, but then you force multiple places changes. 

Maybe the problem here is the code centric approach, if we move to a description first approach, with a sort of validator that, for example,  checks for duplicate position in the same order maybe this can help. Then this file will be use to build the build the `ConfigDef` in the `config()`. Not still sure is feasible, but at least in this way there we be a specific place where configuration should be. The problem is that we can do this for our connectors, transforms etc but the Kafka Connect stuff will remain out of this. Another issue could be easily reusing of common configs. 

Regarding the Localization I think this should be just UI related. What we define in the configuration/field should be in English and the the UI should translate this. Isn't it? @indraraj WDYT about this point?

Naros:
Hi @mfvitale, I understand that the group stuff exists in the Kafka Connect API in `ConfigKeyInfo`; however, I am unaware of anything that forces us to use it. Looking at many of the Kafka Connect transforms, all their field definitions are always injected into the `ConfigDef` with a `null` group and `-1` as the group order.

But I did take a look at the Confluent JDBC code base, and I noticed an interesting approach, see [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java#L449).  They allow the order to be defined when the `ConfigDef` is constructed :thinking:

mfvitale:
> But I did take a look at the Confluent JDBC code base, and I noticed an interesting approach, see [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java#L449). They allow the order to be defined when the `ConfigDef` is constructed 🤔

This is for sure an interesting approach. Effectively you decide the order in the code and the `order` property is automatically adjusted.

jpechane:
@mfvitale One question for you, we already have `io.debezium.config.ConfigDefinition` in place for connectors together with `io.debezium.config.Field`. If we turn `io.debezium.config.ConfigDefinition` to be not connector specific then I can imagine that the classes can expose Debezium's `ConfigDefinition` which would make them independent of Kafka Connect?
In fact at least for connectors we use Debezium's builidng blocke before turning the into Connect ones.
Thinking about it the platform should understand both Debezium and COnnect configs. Dbezium's for decoupling and Connect's for for example 3rd party SMTs.

indraraj:
@mfvitale  An ideal case for supporting localization would be backend returning translation keys instead of actual text. But we can manage without relying on the backend by handling it entirely on the UI side also with mapping the API returned 'en' text to UI translation text keys

mfvitale:
> Thinking about it the platform should understand both Debezium and COnnect configs. Dbezium's for decoupling and Connect's for for example 3rd party SMTs.

@jpechane this is exactly the point the we need to decide. Once decided, we can decide the best approach. 

In my opinion we should support both. Based on that,  we can say:

### Debezium components:

**Discoverability**: via the `ConnectorMetadataProvider` that needs to be changed to be component agnostic
**Configuration discoverability**: via the `io.debezium.config.ConfigDefinition` ans so `io.debezium.config.Field`.

### Kafka Connect components:
**Discoverability**: via the different interfaces that exposes the `org.apache.kafka.common.config.ConfigDef`, in the first phase. Provide a KIP to have an unified interface. 
**Configuration discoverability**: via the `org.apache.kafka.common.config.ConfigKey`

jcechace:
Ideally for our components, I would love to se us turning this on the head and define configuration in declaratively (ie via yaml / json) and then generate the code config definition. However at the same time, I can't imagine how we would make such a change.

So what Jiri and Mario propose actually makes sense to me. What I would also like to see for this descriptor is some sort of (json) schema (meaning schema for the descriptor itself, not the actual instances).

mfvitale:
> What I would also like to see for this descriptor is some sort of (json) schema (meaning schema for the descriptor itself, not the actual instances).

Sure, I'll add it. 

> Ideally for our components, I would love to se us turning this on the head and define configuration in declaratively (ie via yaml / json) and then generate the code config definition.

Yes, this is what I called `descriptor first approach` that for sure will remove lot of complexity in the `*Config.java` classes. 
The main issue I see here is the sharing of common properties. Furthermore, though this is related, it is a big change that maybe needs to be considered in a separate discussion. No blocks us to adopt this approach in the future.

DDD-13.md
```diff
@@ -0,0 +1,529 @@
...
+   2. Modify the assembly to copy generated files to the distribution package
+4. Implement the descriptor service
+5. Support volumes in the Debezium Platform Helm chart
+
+> **_Note:_** Point 2 can be postponed to the end so that we can go thought the whole pieces having only the connector descriptors and then add the others. 
+
+## Open points
+
+* Contribute to adding deprecation/alias support to `ConfigKey`?
+* Support for Storage? Maybe in the next phase?
```
jpechane: Support for storages can come in the next stage.

DDD-13.md
```diff
@@ -0,0 +1,529 @@
...
+   1. Add `debezium-schema-generator` plugin to `debezium-server-dist`
+   2. Modify the assembly to copy generated files to the distribution package
+4. Implement the descriptor service
+5. Support volumes in the Debezium Platform Helm chart
+
+> **_Note:_** Point 2 can be postponed to the end so that we can go thought the whole pieces having only the connector descriptors and then add the others. 
+
+## Open points
+
+* Contribute to adding deprecation/alias support to `ConfigKey`?
```
jpechane: I'd definitely try to contribute as it is something that the whole community can befenfit from but don't rely on it at east now as the lead time in Kafka can be quite slow.

jpechane:
@mfvitale To add a bit more complication to this PR. There was a discussion in https://debezium.zulipchat.com/#narrow/channel/302533-dev/topic/Oracle.20adapter-centric.20refactoring that requires a bit more flexible confiugration for Oracle config. If we are going to support both configurations approaches (and in fact it is really the best way forward) there is `io.debezium.config.Field.withDependents(String...)` method. What if we provide additional one where you could define dependents based on the field value like `io.debezium.config.Field.withDependents(String fieldValue, String... dependents). This way it would be possible for example for Oracle connector to express to UI which fields are/should be used based on selected adapter. @Naros Any comment?

rk3rn3r:
> Ideally for our components, I would love to se us turning this on the head and define configuration in declaratively (ie via yaml / json) and then generate the code config definition. However at the same time, I can't imagine how we would make such a change.
> 

This is imho the best approach, but only achievable for Debezium. We can generate config classes from the descriptors, that'll be doable. There was also a workaround in place to fill missing metadata for Kafka Connect SMTs, Predicates, other connectors, etc for the old UI. We'd need something like this in the descriptor-first implementation too. Somehow a meta-descriptor that adds missing data to those non-Debezium entities. This can also be done with meta-descriptors, which don't generate Config classes but descriptors for those non-Debezium connectors and SMTs, etc.
Does this make sense?

Naros:
> If we are going to support both configurations approaches (and in fact it is really the best way forward) there is io.debezium.config.Field.withDependents(String...) method. What if we provide additional one where you could define dependents based on the field value like `io.debezium.config.Field.withDependents(String fieldValue, String... dependents). This way it would be possible for example for Oracle connector to express to UI which fields are/should be used based on selected adapter.@Naros Any comment?

If I understand you correctly, that should work @jpechane. We must ensure that the UI can support a multi-level hierarchy of dependencies between fields.

For example:

* Setting the connector's connection adapter: LogMiner, LogMiner Unbuffered, XStream, OpenLogReplicator
  * Setting adapter to LogMiner, the buffer type must be set: Memory, Infinispan Embedded, Infinispan Remote, Ehcache
    * Setting buffer type to Infinispan toggles Infinispan-specific fields.
    * Setting buffer type to Ehcache toggles Ehcache-specific fields.

This is definitely expressable from the method you suggested, but I'm not sure how easy this will be from Indra's point of view in the UI, but maybe there are building blocks for this sort of thing already :crossed_fingers:

jpechane:
@indraraj Could you please take a look at Chris' comment if it is/will be achievable in UI?

indraraj:
@Naros, I believe what you are describing can be implemented as nested form fields with interdependency. I guess this should be achievable in UI with form libraries like Formik.

mfvitale:
> > Ideally for our components, I would love to se us turning this on the head and define configuration in declaratively (ie via yaml / json) and then generate the code config definition. However at the same time, I can't imagine how we would make such a change.
> 
> This is imho the best approach, but only achievable for Debezium. We can generate config classes from the descriptors, that'll be doable. There was also a workaround in place to fill missing metadata for Kafka Connect SMTs, Predicates, other connectors, etc for the old UI. We'd need something like this in the descriptor-first implementation too. Somehow a meta-descriptor that adds missing data to those non-Debezium entities. This can also be done with meta-descriptors, which don't generate Config classes but descriptors for those non-Debezium connectors and SMTs, etc. Does this make sense?

It make sense @rk3rn3r but seems that we would like to maintain the code first approach. 

Maybe we can use this to generate the full descriptors with all requirements and then, once we have them, we can think to switch to the declarative approach. WDYT?

DDD-13.md
```diff
@@ -0,0 +1,595 @@
...
+        "type": "string",
+        "display": {
+           "label": "Connector adapter",
+           "description": "The adapter to use when capturing changes from the database.",
+           "group": "Connection advanced",
+           "groupOrder": 1,
+           "width": "medium",
+           "importance": "high"
+        },
+        "valueDependants": [
```
mfvitale, line 89: While creating this example seems that the concept of `dependant` has more sense only related to a `value` and not to a field itself. 

If I'm correct, in the current implementation the dependents declared with `public Field withDependents(String... dependents)` will be used in the `Field.validate(...)` to validate all dependents fields. 

So suppose I have a property `prop1` with `prop2` as dependent, validating the `prop1` will trigger the validation of `prop2`. Should this happen in any case?

So to me seems to have not so much sense. @jpechane @Naros WDYT?

DDD-13.md
```diff
@@ -0,0 +1,595 @@
...
+              "dependants": [
+                 "log.mining.buffer.ehcache.global.config",
+                 "log.mining.buffer.ehcache.transactions.config",
+                 "log.mining.buffer.ehcache.processedtransactions.config",
+                 "log.mining.buffer.ehcache.schemachanges.config",
+                 "log.mining.buffer.ehcache.events.config"
+              ]
+           },
+           {
+              "values": ["infinispan_embedded", "infinispan_remote"],
```
mfvitale, line 126: Seems to me that we have the same kink of information just described in different ways and `dependents` seems to be another one 

We have:
* the `group` that should group similar configuration, why not us it for example for grouping all `adapter` configs?
* the prefixes, for example `log.miner.*` is a kind of groups and also creating "dependencies". Isn't it?
* dependents that create a relation between configs, isn't this doable with `prefixes` or `groups`?

@jpechane @Naros WDYT?

DDD-13.md
```diff
@@ -0,0 +1,595 @@
...
+              "dependants": [
+                 "log.mining.buffer.infinispan.cache.global",
+                 "log.mining.buffer.infinispan.cache.transactions",
+                 "log.mining.buffer.ehcache.processedtransactions.config",
```
mfvitale:
```suggestion
                 "log.mining.buffer.infinispan.cache.processed_transactions",
```

DDD-13.md
jpechane: So do you propose to keep validation only based on the value of the field but not on the field itself?

mfvitale: No, what I'm trying to say is that the `withDependents(String... dependents)` seems not to be so useful while instead `io.debezium.config.Field.withDependents(String fieldValue, String... dependents)` this is the one we really want. 

I don't see any useful usage, as of now, of the `withDependents(String... dependents)`. Maybe I'm missing something.

rk3rn3r:
> Maybe we can use this to generate the full descriptors with all requirements and then, once we have them, we can think to switch to the declarative approach. WDYT?

That brings up 2 questions: which is going to be the source of truth and can the current code represent what we need. Based on the experiences from the legacy UI and @Naros valid concern on the previous implementation I would say no. The good programmatic way to express those things (like dependents or validators) that's very hard to transform. The suggestion for dependents for example is nothing else than changing it to a declarative config. Then we could move towards a full declarative config and code generation. And as we have already a declarative approach from the old connector schemas, we could start from there. But sure, it increases the required effort.

This also brings me back to a point from the past for field dependencies and connector "flavors". The good thing of the declarative approach is that you can basically very easily model those dependencies. You create one schema per connector flavor. This can be hierarchically and based on the selected values you jump from schema "a" (aka "a1") to a2. That way you have SOC and an easy way to model dependencies.

Solving this in code can be hard, especially when it needs to be transformed into something declarative that's usable from a UI or an API.

I hope that makes sense?

mfvitale:
So seems that "declarative approach" has been mentioned multiple times. 

I think we should bring up again the discussion and decide for "declarative" or "programmatic" approach first.

mfvitale:
This now solves debezium/dbz#1204. 

I incorporated the suggestions and I reviewed a bit how the descriptor will be distributed and served. 

Regarding the declarative approach, I think that at certain point we should go for it, but for now I prefer to go with the generation, see how it goes and when we have a stable descriptor we can think to switch and do the opposite. 

@jpechane, @Naros , @rk3rn3r, @kmos feedback are welcomed. I would like to stabilize this so I can start implementing it.

Naros:
I do have one ask: could we find a better way to define `positionInGroup` in the `Field` configuration API? 

For connectors with a small number of configurations, I think defining the ordinal position makes sense, but as the size of the configuration grows, and new features are added that ideally should be logically grouped with other options, this really begins to be a nightmare to manage, particularly because we have no real way to easily visualize all the config options (other than how they're organized in code) and where they show up in the UI relative to others.

I really think we should re-evaluate this comment
https://github.com/debezium/debezium-design-documents/pull/15#issuecomment-2797852288

It wouldn't necessarily need to be in this proper, but I'd like us to think about how we can address this in the near future so that configuration management for connectors doesn't keep getting more and more painful.

mfvitale:
@Naros I din't forgot about this topic. 

> particularly because we have no real way to easily visualize all the config options (other than how they're organized in code) and where they show up in the UI relative to others.

My idea is that the UI could be a good tool to let us to visualize, both position and groups, and discuss on it. Once we define the correct order and grouping we can think to use the approach used [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java#L449)

> but I'd like us to think about how we can address this in the near future so that configuration management for connectors doesn't keep getting more and more painful.

Then as a next step, after we will have a generated JSON, we can think to switch to a declarative approach and use that JSON as source of truth and then generate the java code from it. 

WDYT?

Naros:
Maybe. I do have some concerns around default values and validation with that approach, but we can discuss that when we get to that point.

kmos:
LGTM. I don’t have enough background to comment on potential pain points, so I’ll rely on others for that but sounds as an important feature for the Platform.

Regarding the declarative approach, I agree to proceed and see how it goes and after a stable version we can switch.

kmos (approved):

DDD-13.md
```diff
@@ -0,0 +1,728 @@
...
+The Debezium platform project was started with the idea to provide a centralized and easy way to manage data pipelines on Kubernetes with Debezium Server.
+In that platform, users can configure different components: sources, sinks, and transformations to build their data pipelines.
+For those familiar with Debezium, these components are highly configurable through a set of configuration properties.
+These properties are declared by the components themselves, and usually a property defines some basic information:
+
+* name
+* description
+* type
+* default value
+* validator
```
rk3rn3r, line 13:
```suggestion
* validators
```

rk3rn3r: In the previous schemas we had:

```
per connector:
* required/mandatory fields
* "x-connector-id" : "postgres"
* "x-version" : "1.6.0-SNAPSHOT"
* "x-className" : "io.debezium.connector.postgresql.PostgresConnector"

per field:
* "title" short name / label (for UI field name usage)
* "description" long docs text
* "default" value
* "enum" for enumeration of the available values (snapshot mode etc)
* "format" aka OpenAPI format field
* "type" compatible OpenAPI type
* "nullable"
* "x-name" DBZ config option name ("snapshot.mode" etc)
* "x-category" option category (CONNECTION, CONNECTION_ADVANCED, CONNECTION_ADVANCED_SSL, FILTERS, CONNECTOR, CONNECTOR_SNAPSHOT, etc)
```

rk3rn3r: Validators are hard to represent in a declarative form, probably we need to model the more complex ones and address them by the name.

DDD-13.md
```diff
@@ -0,0 +1,728 @@
...
+      ]
+    },
+    {
+      "name": "table.include.list",
+      "type": "list",
+      "display": {
+        "label": "Include Tables",
+        "description": "The tables for which changes are to be captured",
+        "group": "Filters",
+        "groupOrder": 2,
```
rk3rn3r, line 67: What about reflecting the order by the order in the descriptor?
That would simplify things and aligns the memory and physical ordering with the parsed order.

DDD-13.md
```diff
@@ -0,0 +1,728 @@
...
+              "dependants": [
+                 "log.mining.buffer.infinispan.cache.global",
+                 "log.mining.buffer.infinispan.cache.transactions",
+                 "log.mining.buffer.infinispan.processedtransactions.config",
+                 "log.mining.buffer.infinispan.cache.schema_changes",
+                 "log.mining.buffer.infinispan.cache.events"
+              ]
+           }
+        ],
+        "validation": [
```
rk3rn3r, line 136: was called `validator` or `validators` above, or is this a different option?

rk3rn3r:
Looking at the example schema, we could potentially bring back the SchemaGenerator? The output looks pretty similar.
Nevermind, I discovered: debezium/dbz#1543 .

DDD-13.md
mfvitale: For now we only support basic validation

 *   regex(String)} - Regex pattern validation
 *   enumValues(List)} - Enum/allowed values validation
 *   min(Number)} - Minimum value validation
 *   max(Number)} - Maximum value validation
 *   range(Number, Number)} - Range validation (min and max)

You can have a look to [Validation](https://github.com/debezium/debezium/blob/main/debezium-schema-generator/src/main/java/io/debezium/schemagenerator/model/debezium/Validation.java) and https://github.com/debezium/debezium/blob/d59c2ecf0ded48dc9762196efa386ffbce19882d/debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java#L133

DDD-13.md
mfvitale: Yes the `groupOrder` is something that we could remove. I'm waiting to have also the front-end implementation to see if this could be useful or not.

DDD-13.md
mfvitale: This is the option name, before it was just mentioned as a generic name

DDD-13.md
rk3rn3r: Sure, this aligns with OpenAPI. But some of our validators are more complex and need to be represented in a declarative way aka. giving them a name and mapping them (probably moving them into methods and map by the method name or giving them an alias with an annotation etc).
But I feel like you have it on your list already?

mfvitale: Yes, this is the start. Then we can enhance it and add feature if are required.

rk3rn3r:
How are DDDs numbered? If this is PR is number 15, it should be DDD-15? Else the correct number can only be set right before merging? That leads to gaps in the DDD numbers but would make the process easier?

jpechane:
I'd say that the final number will/should be assigned during the final merge. I don't think we should tie DDD number to PR number.

mfvitale:
> I'd say that the final number will/should be assigned during the final merge. I don't think we should tie DDD number to PR number.

I usually did it in this way. Just coming to my mind a possible glitch on this approach. Referencing the DDD during discussions (i.e. zulip) or in a PR while the DDD is not stabilized and merged could lead to a stale reference then.

Naros:
What we ultimately need is a way to get preassigned numbers that avoid conflicts. We already have this on GH with the issue/pr shared number range, and given we have issues enabled in this repository, what if the process is as simple as opening a GH issue first to be assigned a unique ID, that is used for the DDD?

In the GH issue in the DDD repo, the title might have been "Component descriptors," with the description being an early draft of the motivation and nothing more. The goal of this step is to get an assigned number.

When it comes time to formalize it into a document to be maintained in the repo, we open a PR that references the issue (e.g., "Fixes #<the assigned DDD issue number>") and uses `DDD-<issue #>` as the key. 

This keeps the process similar to what we already do for bugs and tasks, and is also contributor-friendly.

What do you all think?

mfvitale:
> In the GH issue in the DDD repo, the title might have been "Component descriptors," with the description being an early draft of the motivation and nothing more. The goal of this step is to get an assigned number.

So issue is opened in the design document repo? I'm fine with this.

jpechane:
@Naros Sounds good!

PR #16, opened 2025-04-09
DBZ-8903 Debezium Engine Quarkus Extension

kmos:
Initial draft to discuss: https://issues.redhat.com/browse/DBZ-8902


DDD-12.md
```diff
@@ -0,0 +1,157 @@
...
+```
+This approach doesn't work out-of-the-box in situations in which [we want to build a native image of the application](https://debezium.io/blog/2025/03/12/superfast-debezium/).
+
+## 1. Two dependency (Debezium quarkus engine and connector separate)
+The first solution is similar to the actual way to use it inside an application beside the fact that the dependencies are wrapped into quarkus extensions like:
+
+```xml
+<dependency>
+    <groupId>io.debezium</groupId>
+    <artifactId>debezium-quarkus-embedded</artifactId>
```
jpechane:
```suggestion
    <artifactId>debezium-quarkus-engine</artifactId>
```

DDD-12.md
```diff
@@ -0,0 +1,157 @@
...
+    <groupId>io.debezium</groupId>
+    <artifactId>debezium-quarkus-connector-mysql</artifactId>
+    <version>${version.debezium}</version>
+</dependency>
+```
+
+![](./DDD-12/s1-dbz-lib.png)
+
+## 2. One dependency (Debezium quarkus engine with Quarkus connector)
+The second solution propose a unique dependency that contains the `engine` and the `connector` like in this way:
```
jpechane: I'd prefer this solution as I'd say that engine without connector does not make much sense.

DDD-12.md
```diff
@@ -0,0 +1,157 @@
...
+
+the configuration property delegated to define the connector class should be unavailable and already defined inside the extension:
+```txt
+connector.class=io.debezium.connector.mysql.MySqlConnector
+```
+
+![](./DDD-12/s2-dbz-lib.png)
+
+## Quarkus Debezium Extension configuration
+The extension should be configurable like the usual one but with some easy-to-use interface:
```
jpechane: Could we also as a stretch goal support/integrate with `quarkus.datasource.` configuration?

jpechane (commented):
@kmos This is good start, thanks!
I'd like to add support for few more features

* listeners for Connector and Completion callbacks
* Debezium CustomConverter annotation and injection into connector config
* Debezium post-process annotation and injection into connector config

DDD-12.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+the configuration property delegated to define the connector class should be unavailable and already defined inside the extension:
+```txt
+connector.class=io.debezium.connector.mysql.MySqlConnector
+```
+
+![](./DDD-12/s2-dbz-lib.png)
+
+## Quarkus Debezium Extension configuration
```
mfvitale, line 44: Can you provide an example configuration for the extension? Those it have a prefix like `quarkus.debezium`?

DDD-12.md
kmos: like reuse this one? https://quarkus.io/guides/datasource#configure-a-jdbc-datasource:

```
quarkus.datasource.db-kind=postgresql 
quarkus.datasource.username=<your username>
quarkus.datasource.password=<your password>
```

jpechane: Yes, exactly. So if application uses database for both writes and capture changes it can use a single configuration.

mfvitale: I'll propose the use of our property for database connection. Users that want to reuse the jdbc datasource can still do it using [property-expressions](https://quarkus.io/guides/config-reference#property-expressions).

In this way we can be flexible and support single/multiple database.

DDD-12.md
kmos: Yes

DDD-12.md
kmos: Would it be feasible to fall back to `quarkus.datasource.*` in case any configuration is missing from the Debezium one?

Naros: The only database I am concerned about regarding this might be MySQL. In our `MySqlConnectionConfiguration`, we set many connection properties in the `URL_PATTERN`, some of which substantially influence the data the connector gets. 

So, we would likely need to either (a) document the options that should be set by the user-provided URL or (b) have a way to marry what we require with what the user has supplied.

DDD-12.md
```diff
@@ -0,0 +1,156 @@
...
+The extensions must be configurable using the properties and yaml like any Quarkus application. The configuration properties available for the debezium engine must be available using a prefix `quarkus.debezium.xxx` like:
+
+```properties
+quarkus.debezium.configuration.offset.storage=org.apache.kafka.connect.storage.MemoryOffsetBackingStore
```
Naros, line 49: I understand this is just an example, but I would prefer we keep property name prefixes short. We have some connectors that already have obnoxiously long names, and I would really dislike seeing :grin:  :
```
quarkus.debezium.configuration.log.mining.buffer.infinispan.cache.global="..."
```
That said, I do think using `quarkus.debezium.*` as a prefix aligns with many of the other Quarkus extensions and that would not clash with the outbox that uses `quarkus.debezium-outbox.*`, but I am curious how will this clash specifically where we want to use this as the basis for Debezium Server that uses the configuration prefix of just `debezium.*` ?  Does DS just map to the extension's configs for the source?

DDD-12.md
```diff
@@ -0,0 +1,156 @@
...
+quarkus.debezium.configuration.name=native
```
Naros: Out of curiosity, what is the purpose of the `native` configuration here? Would we not have the same configuration and simply compile as JVM vs Native modes?

DDD-12.md
```diff
@@ -0,0 +1,156 @@
...
+## Quarkus Debezium Extension additional feature
+
+The extension permits to address some use-cases already present in Debezium but in a _Quarkus_ way:
+
+- `Debezium Listener`
+- `Custom Debezium Converter`
+
+### Quarkus  Debezium Listener
+
+a Quarkus Developer using a `Debezium Listener`  can intercept events (`INSERT, UPDATE, DELETE...`) from a table like `order`, with a simple annotation like:
```
Naros, line 179: I could also see there being additional listeners for things like (which I just saw Jiri mentioned a few of these above and I missed it);

- Annotated handlers for notification events, snapshot started, skipped, completed, etc.
- Annotated handlers for schema changes
- Annotated handlers for heartbeats
- Lifecycle events, i.e. configured, started, stopped, failed.

As a stretch goal, could we also consider favoring annotation over configuration? 

For example, if all the listeners define they're only interested in `INSERT` events, then we could specifically deduce that the configuration should have an implicit `skipped.operations` that skips everything but `c` events.  The same could hold for things like `table.include.list`.

Another example would be integrating the service registry with Quarkus CDI so that Debezium-created objects could inject CDI beans into them inside a Quarkus application.

I'd also consider providing annotated ways to register things like a `CustomConverter`, `PostProcessor`, and `TopicNamingStrategy`, just to name a few. Again, favoring annotation over configuration.

Naros (commented):
@kmos I agree with all the feedback, an excellent start :tada: 

I've added some more ideas below.  They're certainly nice to haves, and I prefer to consider everything, including the kitchen sink, so we have that long-term idea so as we walk through this we already have ideas about phase 2, 3 and so on to avoid making an early poor decision.

Naros:
@jpechane, with this eventually being the basis for Debezium Server and the DebeziumListener idea being the equivalent to the DS change event consumer, should we consider the possibility of leaving the door open for multiplexing events to different consumers?  Given that the basis lends itself to the fact there could be different listeners defined in the application, technically there would be nothing stopping that in DS either.

DDD-12.md
kmos: No no, it's only a copy-paste from the quarkus native [example](https://github.com/debezium/debezium-examples/blob/331566353f2e3c7d7f00f772c0b5791300a48b2a/quarkus-native/src/main/resources/application.properties#L4). There is no correlation with the compilation mode.

jpechane:
@Naros I think this is viable. But we really need to be very careful with the contract and implementation then. We'd need to make sure that offsets are flushed only if all listeners comitted it.
In this regard I wonder if we could just stay with what we have and then we can provide an aggregate listener (composite design pattern) that would be used to manage mutliple regular listeners.

DDD-12.md
kmos: > we also consider favoring annotation over configuration

I perfectly agree! I think that should be a goal/principle of the extension. It should be something to take in mind during the evolution of the extension.

> Another example would be integrating the service registry with Quarkus CDI so that Debezium-created objects could inject CDI beans into them inside a Quarkus application.

Yeah, for sure the service registry integrated with the CDI can help also the native build.

DDD-12.md
kmos: I agree that we need (unfortunately) to keep aligned to other extensions (`quarkus.extension-name`) and dbz-server (`debezium.server`).

> Does DS just map to the extension's configs for the source?

I guess so. In the case of Debezium Server (DS), we will use the configuration prefix `debezium.*`, which maps to the extension (`quarkus.debezium.*`)

DDD-12.md
```diff
@@ -0,0 +1,326 @@
...
+
+public class OrderDeserializer extends ObjectMapperDeserializer<Order> {
+    public OrderDeserializer() {
+        super(Order.class);
+    }
+}
+```
+
+```properties
+quarkus.debezium.deserializer=com.acme.order.jackson.OrderDeserializer
```
mfvitale, line 269: How I can provide multiple deserializers?

DDD-12.md
```diff
@@ -0,0 +1,326 @@
...
+
+```java
+import io.debezium.engine.InsertEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped  
+class CustomPostProcessor {  
+  
+    @DebeziumPostProcessor()  
+    public void process(Object key, Struct value) {
```
mfvitale, line 317: There should be a return type here. Isn't it?

DDD-12.md
```diff
@@ -0,0 +1,326 @@
...
+class SchemaChangeListener {  
+  
+    @DebeziumSchemaChangeListener()  
+    public void listener(SchemaChangeEvent event) {  
+        /// some logic to apply
+    }  
+}
+```
+
+### Quarkus Debezium Snapshot Listener
```
mfvitale: Maybe here we should generically caputure the [Notifications](https://debezium.io/documentation/reference/3.1/configuration/notification.html) here. WDYT?

DDD-12.md
kmos: At this stage, I didn't think I would have to support it. If I have to answer (and support it), similar to the [quarkus kafka library](https://quarkus.io/guides/kafka#serialization-autodetection), we can instrument the listener with something like that:

```java
import io.debezium.engine.InsertEvent;
import jakarta.enterprise.context.ApplicationScoped;

@ApplicationScoped  
class OrderListener {  

    @DebeziumListener(OrderDeserializer.class)  
    public void listener(InsertEvent<String, Order> event) {  
        /// some logic to apply
    }  
}
```
or

```java
@ApplicationScoped  
class OrderListener {  

    @DebeziumListener("order")  
    public void listener(InsertEvent<String, Order> event) {  
        /// some logic to apply
    }  
}
```

```properties
quarkus.debezium.order.deserializer=com.acme.order.jackson.OrderDeserializer
```

DDD-12.md
jpechane: One point WRT annotations - are we going to annotte our OOTB  custom converteres and post processors? If we do we need to put the annotation into Debezium API module so core does not depend on Quarkus.

kmos: > are we going to annotte our OOTB custom converteres and post processors?

No, I would avoid (if possible) to annotate the OOTB converters like `ReselectColumnsPostProcessor`.

DDD-12.md
kmos: Well actually the interface is written in this [way](https://github.com/debezium/debezium/blob/6a37d34f627f878280f692d51e498c1c64840605/debezium-core/src/main/java/io/debezium/processors/spi/PostProcessor.java#L30).

and the comment:

>  * While this may seem similar to a Kafka Transformation, the difference is that a post processor
>* operates on the raw {@link Struct} objects which are still mutable, allowing for a variety of
>  * use cases that wish to have the full data set of a row for the post-processing step.

IMHO, it's really powerful, and I'm not sure I’d want to give that much power to a *quarkus developer*. On the other hand, I’m not sure how to express it differently

kmos: Do you have any suggestion?

mfvitale: @Naros Any particular reason why this was not something like `public Struct process(Object key, Struct value)` so letting the implementors to provide the post processed `Struct`?

DDD-12.md
kmos: Well, if you are talking about `Custom Notification` (since sink, log, and JMX are already tailored to specific use cases), it depends. Providing access to the `ChangeEvent` is certainly [powerful](https://github.com/debezium/debezium/blob/6a37d34f627f878280f692d51e498c1c64840605/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java#L224), but on the other hand, the [NotificationChannel](https://github.com/debezium/debezium/blob/6a37d34f627f878280f692d51e498c1c64840605/debezium-core/src/main/java/io/debezium/pipeline/notification/channels/NotificationChannel.java#L22) doesn't seem powerful enough to support custom deserializers/converters.

mfvitale: No, I'm talking more about something like https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java but for the Notifications. 

So we should provide a `in-process` notification channel and then the user can get can just start listening to those notifications in a provided listener.

DDD-12.md
Naros: @mfvitale the main reason was to enforce the idea that you shouldn't change the `Struct` schema, that you should only be mutating the field values. 

In some places where the `EventDispatcher` calls into the `PostProcessor`, the envelope's schema has been precomputed, so changing the value's `Struct` schema layout would require a recreation of the envelope schema. At the time we added this, we wanted to be as surgical and precise as possible without introducing too much risk as it was a last minute addition.

Naros: I am not sure that we want to be in the habit of where we need to re-create the envelope schema anytime a user uses a `PostProcessor`, as the entire point of the `TableSchemaBuilder#create` method was to do this once everytime a table is added or altered.

That said, I can see the power that allowing for this could have.  @jpechane any thoughts?

jpechane: I'd keep the things as they are now - so no change in schema. We can think about opening this option in future but for now I'd like to have the extension provide the same feature set we have in Debezium core. Let's avoid scope creep.

DDD-12.md
kmos: If I understand correctly, what do you mean is to have an annotation that can access to  notifications and hook particular notifications (like snapshot)? like:

```java
@ApplicationScoped  
class NotificationChannel {  

    @DebeziumNotificationHandler(SnapshotNotification.class)  
    public void handler(Notification notification) {
        /// some logic to apply
    }
}
```

Could you please provide an example?

DDD-12.md
kmos: Maybe documentation is the quickest thing we can do.

DDD-12.md
mfvitale: @Naros sounds good. Let's keep it as is.

DDD-12.md
mfvitale: For now I think just having a way to listen to all notifications will be fine.
```java
@ApplicationScoped  
class NotificationChannel {  

    public void handler(Notification notification) {
        /// some logic to apply
    }
}
```

mfvitale:
Great work. LGTM.

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+
+```java
+import io.debezium.engine.InsertEvent;
+import jakarta.enterprise.context.ApplicationScoped;  
+import io.debezium.engine.quarkus.Operation.INSERT;
+
+@ApplicationScoped  
+class OrderListener {
+  
+    @DebeziumBatchListener("order", INSERT)
```
jpechane: How are you going to express stituation when you want to listen for multiple tables and multiple event types in the same method?

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+```java
+import io.debezium.engine.ChangeEvent;
+import jakarta.enterprise.context.ApplicationScoped;  
+
+
+@ApplicationScoped  
+class OrderListener {
+  
+    @DebeziumListener("order")  
+    public void listener(ChangeEvent<String, String> event) {
```
jpechane: Are we going also support exposition of raw `SourceRecord` object?

DDD-12.md
kmos: we can use an array of strings like:

```java
@DebeziumBatchListener("order") // works because Java auto-wraps into array
public void listener() {}

@DebeziumBatchListener({"order", "product"})
public void listener {}
```

kmos: in the same way for the event types

DDD-12.md
kmos: Probably we should support it. WDYT?

DDD-12.md
jpechane: What if we'd split it into multiple annotations?
`@DebeziumBatchListener` would match everything. Then there would be repeatable annotation `@Collection` or similar where you define the source of the event and another set of annotations `@Create/@Update/@Delete` that would alow you to specify type of events?

DDD-12.md
jpechane: I'd say so, it is one of the things that could be useful for Debezium Server.

DDD-12.md
Naros: I think if we move to the composition style of annotations, then perhaps we consider separating these based on target/type versus method. For example, the `@DebeziumBatchListener` could be defined at the class level while we allow `@Collection` and the crud-specific ones as method-only annotations.

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+
+The Quarkus Debezium extension allows you to be notified of the engine's state using the following annotation:
+
+```java
+import io.debezium.engine.source;
+import jakarta.enterprise.context.ApplicationScoped;
+
+
+@ApplicationScoped
+class DebeziumEngineLifeCycle {
```
Naros, line 143: I think it would be far more idiomatic and more Quarkus / CDI driven if we instead relied on CDI events paired with the `@Observes` annotation for these types of callbacks rather than rolling custom annotations. The CDI event also then doubles as a parameter object, allowing us to pass an unlimited amount of context to each callback that we believe makes sense. wdyt?

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+```java
+import io.debezium.engine.ChangeEvent;
+import jakarta.enterprise.context.ApplicationScoped;  
+
+
+@ApplicationScoped  
+class HeartbeatListener {
+  
+    @DebeziumHeartbeat()  
+    public void heartbeat(ChangeEvent<String, String> event) {
```
Naros: I also think for heartbeat events, we could again rely on `@Observes DebeziumHeartbeat` that provides access to the event along with any other heartbeat contextual details we may want to provide the handler.  wdyt?

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+
+```java
+import io.debezium.engine.ChangeEvent;
+import jakarta.enterprise.context.ApplicationScoped;  
+
+
+@ApplicationScoped  
+class HeartbeatListener {
+  
+    @DebeziumHeartbeat(query="SELECT now()", interval=10000)
```
Naros: Let's say I want to create a reusable heartbeat handler that sends events to a Kafka topic or another target type, then I should be able to define that class with a method that `@Observes DebeziumHeartbeatEvent`. The point of this class is not to define configuration, but to define the handler logic irrespective of how the heartbeat is triggered. This allows maximum reuse. If we mix the configuration concern here, I find that's breaking SoC.  wdyt?

DDD-12.md
Naros: So in hindsight, circling back to my recent theme on CDI events, I wonder if we're getting too much into the weeds even with annotations here. I wonder if it would make more sense to simply have something like:

```java
public void listener(@Observes DebeziumChangeEvent event) {
  if (DebeziumChangeEvent.Type.INSERT.equals(event.getEventType()) {
    /* handle insert */
  }
  if ("my-topic-name".equals(event.getDestination()) {
    /* handle event for my-topic-name
  }
}
```

That's not to say that we couldn't still have the `@Collection` or CRUD annotations to do engine-side filtering before dispatching the event, but this would eliminate the listener-centric annotations.

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+
+```java
+import io.debezium.engine.InsertEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped  
+class SchemaChangeListener {  
+  
+    @DebeziumSchemaChangeListener()  
+    public void listener(SchemaChangeEvent event) {
```
Naros:
```suggestion
    public void listener(@Observes DebeziumSchemaChangeEvent event) {  
```

DDD-12.md
```diff
@@ -0,0 +1,322 @@
...
+
+```java
+import io.debezium.engine.InsertEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped  
+class SnapshotListener {  
+  
+    @DebeziumNotificationHandler()  
+    public void handler(Notification notification) {
```
Naros:
```suggestion
    public void handler(@Observes DebeziumNotificationEvent notification) {
```

Naros (commented):
I left a few inline comments wrt using CDI `@Observes` in favor of a plethora of custom annotations. Not sure what everyone's take is on this, but it feels more Quarkus-like to me personally.

DDD-12.md
kmos: I think you're absolutely right. I definitely prefer using the `@Observes` annotation.

DDD-12.md
jpechane: @Naros  In this case maybe qualifier annotations could be used to implement filtering for events?

Naros: Exactly @jpechane :+1:

kmos: > What if we'd split it into multiple annotations?
@DebeziumBatchListener would match everything. Then there would be repeatable annotation @Collection or similar where you define the source of the event and another set of annotations @Create/@Update/@Delete that would alow you to specify type of events?

It really depends on the kind of experience we want to offer to developers. If the goal is to provide something in the style of an _ORM_, then I believe your proposal makes perfect sense. On the other hand, if we're aiming for more of an event streaming approach, concepts like `Listener`, `Consumer`, and `Producer` might feel like a more natural fit.

> So in hindsight, circling back to my recent theme on CDI events, I wonder if we're getting too much into the weeds even with annotations here. I wonder if it would make more sense to simply have something like: [...]

The `@Observes` annotation, while it fits quite well for `heartbeat events`, may sound a bit different compared to conventions used in other frameworks or libraries. Just to give you a few references of what I mean:

- `@Incoming("xxx")` for [quarkus kafka extension](https://quarkus.io/guides/kafka#receiving-messages-from-kafka)
- `@KafkaListener()` for [spring-kafka](https://docs.spring.io/spring-kafka/reference/kafka/receiving-messages/listener-annotation.html#record-listener)

DDD-12.md
kmos: Totally agree

DDD-12.md
kmos: I totally agree again. The idea here is to have in the listener the reason _why_ the function is invoked (_the sql query_).

DDD-12.md
kmos: As I mentioned [here](https://github.com/debezium/debezium-design-documents/pull/16#discussion_r2079886739), `Observes` sounds more like something you’ve _seen_ without taking any action. On the other hand, terms like `listener` or `consumer` suggest that some kind of interaction has occurred — such as a commit — which might better reflect the intended behavior.

DDD-12.md
kmos: totally agree!

DDD-12.md
jpechane: @kmos If you refere to Quarkus kafka extension then maybe the annotation should be called `Captures` or `Capturing`?

kmos: `Capturing` sounds really good for me

kmos: to recap a bit:

The `capturing` should intercept the `captures` coming from a source in which Debezium is configured (sql, no-sql). The following approach:

```java
@ApplicationScoped
public class CaptureHandler {

    @Capturing
    public void capture(RecordChangeEvent<SourceRecord> event) {
        // apply your logic
    }
}
```

corresponds to have a `Consumer` in the debezium engine:

```java
DebeziumEngine.create(ChangeEventFormat.of(Connect.class))
                .using(config.asProperties())
                .notifying((record) -> {
                    // your logic
                })
                .build()
```

we could have a series of filtering options instrumented in the annotation (based on the most important information). The point is what are the concepts that can fit here? @Naros [point us](https://github.com/debezium/debezium/pull/6426#discussion_r2095653050) to the possibility to have some drawbacks if we use the [FQN](https://debezium.io/documentation/reference/3.1/configuration/signalling.html#debezium-signaling-enabling-source-signaling-channel) (*fully qualified name*) in cases in which the source is no-sql and suggest to be more *topic oriented*. In my mind the topic is a concept driven by predicates like `TopicNameMatches`.

DDD-12.md
```diff
@@ -0,0 +1,349 @@
...
+ * <p>
+ * The Engine is submitted to an {@link Executor} or {@link ExecutorService} for execution by a single thread
+ */
+public interface Debezium {
+
+    /**
+     * @return engine's signaler, if it supports signaling
+     * @throws UnsupportedOperationException if signaling is not supported by this engine
+     */
+    Signaler signaler();
```
jpechane, line 94: @kmos I wonder if we could expose Signaller also via CDI events. So instead of direct access to signaller a signal event would be emitted.

kmos: Something like:

```java
import io.debezium.engine.InsertEvent;
import jakarta.enterprise.context.ApplicationScoped;

@ApplicationScoped  
class Signaler {  
    
    public void handler(@Observes Signal signal) {
        /// some logic to apply
    }
    
}
```

@jpechane ?

mfvitale: I think @jpechane meant a way to `produce` signals.

kmos: ok so something like:

```java
import io.debezium.engine.InsertEvent;
import jakarta.enterprise.context.ApplicationScoped;

@ApplicationScoped  
class SomethingThatWantToSignal {
   @Inject
   Signaler signaler;
    
    public void produce() {
        /// some logic
       signaler.signal(new DebeziumEngine.Signal("1", "log", "{\"message\": \"Signal message at offset ''{}''\"}", null));
    }
    
}
```

DDD-12.md
kmos: as mentioned [here](https://github.com/debezium/debezium/pull/6426#discussion_r2098488145):

> I think that we can evaluate to use concept like `topic` and `RecordFilterStrategy` which is a `predicate` of `ConsumerRecord<K,V> -> boolean` and use a similar approach. What I am wondering is there any generic *identifier* that we can use for the `source` that isn't strictly related to the Kafka domain?

I propose also a `ChangeDataFilterStrategy` to use as filter for the events

DDD-12.md
jpechane: I aimed at something like https://quarkus.io/guides/cdi#events-and-observers if possible.

kmos: something in the *quarkus way* to produce event:

```java
import io.debezium.engine.InsertEvent;
import jakarta.enterprise.context.ApplicationScoped;

@ApplicationScoped  
class SomethingThatWantToSignal {
   @Inject
   Event<Signal> event;
    
    public void produce() {
        /// some logic
       event.fire(new DebeziumEngine.Signal("1", "log", "{\"message\": \"Signal message at offset ''{}''\"}", null));
    }
    
}
```

DDD-12.md
kmos: Now that I am going deep on Debezium internals, I was thinking that the *Filter* use case can be addressed with `Predicates`. I was thinking to introduce SMT and predicates with the same approach of this https://github.com/debezium/debezium/pull/6526.

jpechane: @kmos I am fine with supporting Predicates but I think we should support our filtering to not so we are not tied fully to kafka Connect concepts.

kmos:
I think that we can merge this PR. We have developed the majority of the features in the document.

PR #17, opened 2025-04-27
DDD-11 Efficient Heap-based Transaction Buffer

Naros:


DDD-11.md
```diff
@@ -0,0 +1,165 @@
+# Universal Transaction Buffer
+
+## Motivation
+
+Most Debezium connectors are provided transaction log changes for committed transactions, meaning that the connector can immediately dispatch changes safely.
+However, for some connectors like Oracle and iSeries, the transaction logs contain a mix of committed, uncommitted, and rolled back changes.
```
jpechane, line 6:
```suggestion
However, for some connectors like Oracle and Informix, the transaction logs contain a mix of committed, uncommitted, and rolled back changes.
```

Also there is https://debezium.io/documentation/reference/3.1/connectors/mysql.html#mysql-property-binlog-buffer-size

DDD-11.md
```diff
@@ -0,0 +1,165 @@
...
+A producer (aka `ExcerptAppender`) writes a document to the queue, and consumers (aka `ExcerptTailer`) consumes changes from the queue.
+
+In addition, the queue framework supports multi-producer / multi-consumer patterns out of the box.
+This means we could easily decouple transaction log reading from event parsing and dispatch.
+
+### Transaction Reader
+
+During the streaming phase, we would start a secondary thread that performs the transaction log reading operations.
+This thread would use the database API to gather all the events, and push the data onto the queue.
+Data is appended to the queue using the `ExcertAppender`, like as follows:
```
jpechane, line 49:
```suggestion
Data is appended to the queue using the `ExcerptAppender`, like as follows:
```

DDD-11.md
```diff
@@ -0,0 +1,165 @@
...
+
+The Chronicle Queue API also supports writing objects that extend `SelfDescribingMarshallable`, allowing for a more object-oriented driven approach.
+In this way, you would choose to deserialize each JDBC row into an object representation and then write that to the queue instead.
+
+When it comes to read position offset management, there are two ways the reader thread could manage this:
+
+* Position state is transient. Upon connector restart, the queue files are removed from disk, and reading resumed from the connector offset position (like today).
+* Position state is persistent, and upon restart, the resume position is based on the last read position the thread captured. If we maintain a second offset value for the reader position, we could choose to fallback to the offset position like we use today if there are no queue files persisted on disk, or resume from the last read position if queue files do. This provides the most efficient management and integration between relying on the persisted queues versus re-reading data that may no longer be available.
```
jpechane, line 100: I'd aim for two stage implementation. The first solution is the most straightforward and will be great in validation the concept. The additional state management should be added only later.

jpechane (commented):
@Naros Thanks for the proposal.
I've few comments - the title is Universal Transaction Buffer but the content is clearly about a concrete implmentation.
I personally would prefer probably two different DDDs or this DDD should be split into two sections
1) There should be an SPI intorduced that will separate transaction buffering from the implementation
2) There will be one cocerete implementation of this SPI

Also I believe there is an overlap with https://issues.redhat.com/browse/DBZ-8968

WRT Chronicle - how heavy this new dependency would be? Are there any transitive dependencies?

vjuranek (approved):

Naros:
@jpechane with the rework, I believe this touches on your two points, as toward the end where I speak about the generalized form, that would be the SPI/common code and then there is the Oracle example using it.

As for the overlap, I didn't consider that in this pass fwiw.

As for why I am reviving this - in debezium/dbz#2504, we need to have a way to apply transaction semantics to the buffer, and while it can be applied to what we have today, it's very error-prone just due to the in-flight manipulation we do during partial rollbacks. 

So the idea is to shift away from in-flight pruning and instead read the transaction as it is in LogMiner, enqueue all relevant events, and only once we have the full transaction do any reconciliation passes. This makes the truncate logic needed for debezium/dbz#2504 significantly easier to manage, and we don't need to be concerned with any sort of question of whether a previous mining pass has or has not removed an earlier event since it's all delayed to the end of the transaction when we see the commit. This also lets us potentially fold buffered and unbuffered more into one another and reduce code complexity.

To support this model, the transaction buffer needs memory optimizations to reduce its footprint, since delaying those pruning passes will slightly increase the memory footprint for partial rollbacks.

DDD-11.md
```diff
@@ -0,0 +1,986 @@
...
+```
+
+The state falls into three groups.
+
+The first group is the immutable transaction metadata, which is the same information the connector tracks about a transaction today.
+The one field worth calling out is `startScn`, which doubles as the anchor that every event's SCN is stored relative to; more on this under [SCN deltas](#scn-deltas).
+
+The second group is the per-event metadata.
+Rather than allocating an object per event, each metadata field has its own array, and the event's id (its position in the transaction, `0` to `eventCount - 1`) is the index into every one of them.
+Reading event `n` means reading `type[n]`, `scnDelta[n]`, and so on.
```
mfvitale, line 92: Just nitpick, Should event `n` be accessed with `type[n-1]`?

DDD-11.md
```diff
@@ -0,0 +1,986 @@
...
+    final BigInteger startScn; // anchor scn all event scn deltas are based
+    final long startTime;
+    final String userName;
+    final String clientId;
+    final int redoThread;
+
+    private int eventCount = 0;
+    
+    // Per-event metadata, ~59 bytes across all arrays
+    private byte[] type = new byte[8];
```
mfvitale, line 66: Any reason why starting with 8? 

Maybe worth to mention here that these arrays are meant to dynamically be resized.

Naros: I chose `8` to be conservative, especially for short-lived transactions. We can certainly seed the arrays with a much larger value if we think we need to. Just for reference, here are some breakpoints:

|Size|Array Bytes|100 concurrent trxs|
|---|---|---|
|8|648|~64KB|
|16|1120|~112KB|
|32|2064|~207KB|
|64|3952|~396KB|
|128|7728|~773KB|
|256|15280|~1.5MB|
|512|30384|~3.1MB|
|1024|60592|~6.1MB|

For short-lived transactions with 1 to 8 events, at 1024, that's about 59KB wasted, or close to 6MB wasted with 100 concurrent transactions. Based on the table, do you have a preference?

mfvitale: @Naros Sorry could you elaborate more on that table. `Size` is clear and `Array Bytes` I suppose is the size of an empty allocated array of bytes, but what about `100 concurrent trxs`?

Naros: The `100 concurrent trxs` is just `100 * Array Bytes`; it's the total number of bytes the event arrays would consume when 100 concurrent transactions overlap. Keep in mind, this is just metadata, and has nothing to do with chunks or the column values serialized into the chunks.

mfvitale: Well then all depends on how many events by average are in a tx. If in your experience they fits the `8` then I think its fine.

Naros: There's really no good one size fits all answer, at least for Oracle. Maybe we roll the dice and pick `32` or `64` and then get feedback on how it works and go from there. :shrug:

DDD-11.md
Naros: Looking back at the current connector source, at least in Oracle terms, no - event ids are 0-based.

mfvitale: so then `the event's id (its position in the transaction, `0` to `eventCount - 1`)` this should change?

PR #19, opened 2025-06-06
DBZ-8968: Add support for pluggable ArrayDeque

adasari:


DDD-10.md
```diff
@@ -0,0 +1,112 @@
+# DBZ-8968: Add support for pluggable ArrayDeque
+
+## Motivation
+
+Debezium (DBZ) connectors use an in-memory queue (ArrayDeque) within the ChangeEventQueue class to buffer changes between the database reader and the Kafka Connect task. When the Connect task (Eg: IcebergConsumer) is slow to process events, the in-memory queue can become full, causing the database reader to slow down due to backpressure.
```
mfvitale, line 5: HI @adasari thanks for this proposal. 

> between the database reader and the Kafka Connect task. 

I want just to clarify that the queue here is just used as a back pressure mechanism between the database and the Kafka broker. 

> When the Connect task (Eg: IcebergConsumer) is slow to process events,

So this has not so much sense here. Isn't it?

adasari: Yes - queue is used as a back pressure mechanism between the database and Kafka Broker/Target. So, it limits the throughput for improved reliability. 

https://issues.redhat.com/browse/DBZ-8968?focusedId=27073116&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27073116

Postgres Connector is memory bound. So, its throughput is limited by the queue capacity. snapshot reads can be parallelized using the `snapshot.max.threads` setting, the benefits are limited without a fast and efficient writer because of limited internal queue capacity.

The goal is to reduce dependency on heap memory by introducing a pluggable queue to use alternative implementations like off-heap, file system, or hybrid queues (e.g., [tiered-queue](https://github.com/adasari/tiered-queue)), which brings several benefits:

1. Cost efficiency – Smaller instances can run connectors without major performance hits.
2. Higher throughput – Local disk I/O (e.g., with embedded persistent queues) is faster than repeated remote DB access in case of back pressure.
3. Better resilience – Settings like max.queue.size.in.bytes help control the load spikes, but affects the performance. 

These options offer flexibility in balancing cost and throughput, and can be valuable for Debezium use cases.

Let me know your thoughts. thanks.

DDD-10.md
```diff
@@ -0,0 +1,112 @@
...
+
+For example, The Postgres Connector is single-threaded and memory-bound, which limits throughput. 
+In snapshot phase, DB reads can be parallelized using the `snapshot.fetch.threads` config. However, the benefits are limited without a fast and efficient writer because of limited internal queue capacity.
+
+This behavior introduces limitations, including:
+1. Memory pressure and frequent garbage collection
+2. OOM risks under high load (max.queue.size.in.bytes introduces additional overhead)
```
rk3rn3r, line 12: Do we really have many reports of OOM issues practically?

Naros: It has the potential, more so in some connectors than others. And as we start to consider introducing a transaction buffer for Informix and PostgreSQL, that potential will increase.

DDD-10.md
```diff
@@ -0,0 +1,112 @@
...
+3. Need for vertical scaling (large heap sizes)
+4. Limited throughput for memory bounded connectors (Eg: PostgreSQL)
+
+## Goals
+
+Introduce a pluggable queue abstraction that allows custom implementations of event buffering for
+1. Cost efficiency : Small instances can be used to host the connectors.
```
rk3rn3r, line 19: Unfortunately you mostly look at Postgres and you seem to look at it's memory footprints. Did you look at other connectors as well?
Especially when I look at connectors supporting schema history like MySQL connector, I assume instance size cannot be reduced as much as you might expect here. If that's true then I am not convinced if externalizing the queue buffer (alone) brings the expected result. wdyt?

jpechane: I'd also like to add that we have (really long standing) plan to address the schema history memory consumption separately - also maybe adding and SPI for implementations that would combine caching with extrenalization or some schema specific re-use of components.
So I'd see this effort as orthogonal to schema history memory issues and I see it valuable too.

adasari: @rk3rn3r You’re right that the cost efficiency may vary across connectors. While memory savings might be more apparent with some, like Postgres, the main goal here is to reduce dependency on in-memory buffering, which helps improve reliability when handling large datasets.

PR #20, opened 2026-01-02
DDD-15 Debezium Server Sink Extension

kmos:


DDD-15.md
```diff
@@ -0,0 +1,124 @@
...
+
+- configuration definition
+- pre-engine sink instrumentation
+- post-engine sink instrumentation
+
+The features should be available in the extensions and in debezium server.
+
+## Debezium Sink Configuration Mapping
+
+Based on [smallrye-config](https://smallrye.io/smallrye-config/3.11.1/config/mappings/), a developer should be able to define mapping rules for the Sink that are compatible with `quarkus` and `debezium server`. It should be available an annotation `DebeziumSinkConfiguration` with a prefix value that define the root mapping of the configuration elements. Based on the case (debezium server or general quarkus extension) the configuration is prefixed with `debezium.sink` or `quarkus.debezium.sink` like:
```
Naros, line 59: @kmos, is this intended to be part of the extension or the rewrite of Debezium Server?

If this is intended for Debezium Server, I fully agree. But if this is designed to be baked into the Quarkus extension, I question if we aren't getting too far into implementation details of what a Quarkus application may do here. The extension should focus on getting data into Quarkus beans for use (annotations for routing those into said beans), and what happens after the bean has it is entirely out of scope imo; unless I have missed some crucial discussion element.

jpechane: I'd probabaly agree here. It seems to me that we are discussing sink specifics. Sink is concept that is distinct for Debezium Server.
I'd proposoe to have an extension defined in the `debezium-server` repository. If and only if we find that at the and the concept can be generalized it would be migrated to the core Quarkus extension.

kmos: No, it isn’t intended to be part of the Quarkus extension. My idea is to make the sink portable across Quarkus projects and Debezium Server without adding anything to the Quarkus engine extension. This would involve creating a set of annotations and APIs to simplify sink development and ensure interoperability. Using this approach should also make it easier to inject configuration into the sink (see the `BeforeCapturing` example).

DDD-15.md
```diff
@@ -0,0 +1,124 @@
...
+
+```java
+package io.debezium.sink.kafka.configuration;
+
+import io.debezium.quarkus.spi.DebeziumSinkConfiguration;
+import io.smallrye.config.WithName;
+
+
+@DebeziumSinkConfiguration("kafka")
```
mfvitale, line 68: It's not cleat to me who manages this annotation? Is something done by the extension? I miss the `@ConfigMapping` here...

kmos: If I remember correctly, you cannot override the prefix (`debezium` or `quarkus`) with `@ConfigMapping` and for this reason I am proposing a custom annotation. It can be processed by another module like `debezium-quarkus-sink`. The important thing is that is processed at build time and substituted with a `@ConfigMapping` with the right prefix (`quarkus.kafka` / `debezium.kafka` )

PR #21, opened 2026-01-13
Quarkus Debezium Cache/Search Invalidation

kmos:


DDD-16.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+### Second Proposal
+
+A developer should import the extension and the relative datasource for cache invalidation:
+
+```xml
+<dependency>
+    <groupId>io.debezium</groupId>
+    <artifactId>debezium-quarkus-cache-invalidation</artifactId>
```
jpechane: cache is a bit generic here, it should be jpa-cache or something like that as it is specific for certain type of scenarios

DDD-16.md
```diff
@@ -0,0 +1,162 @@
...
+
+        @Capturing()
+        public void evict(CapturingEvent<SourceRecord> record) {
+            if (registry.get(record.topic())) {
+                Long itemId = ((Struct) record.key()).getInt64(registry.get(record.topic()).key());
+                Struct payload = (Struct) record.value();
+                Operation op = Operation.forCode(payload.getString("op"));
+
+                if (op == Operation.UPDATE || op == Operation.DELETE) {
+                    emf.getCache().evict(Item.class, itemId);
```
jpechane, line 122: Can we update the cache instead of doing just eviction? Or this can be done via cache provider only?

DDD-16.md
```diff
@@ -0,0 +1,162 @@
...
+
+- no snapshot
+- in memory Schema Changes
+- all configuration defaulted and named `invalidation`
+
+## Behaviour
+
+### Hibernate Cache Eviction
+
+The extension should scan for JPA/Hibernate metamodel at startup and identify all the entities that are under L2C (`Cacheable`) and register a generic `@Capturing` handler delegated to evict the cache. Here a snippet that explain the idea:
```
jpechane: Could you also evaluate Hibernate specific implementation that would be implemented via Debezium-enhanced cache provider?

DDD-16.md
```diff
@@ -0,0 +1,162 @@
...
+        public void evict(CapturingEvent<SourceRecord> record) {
+            SearchSession searchSession = Search.session(em);
+
+            if (registry.get(record.topic())) {
+                Long itemId = ((Struct) record.key()).getInt64(registry.get(record.topic()).key());
+                Struct payload = (Struct) record.value();
+                Operation op = Operation.forCode(payload.getString("op"));
+
+                if (op == Operation.UPDATE || op == Operation.DELETE) {
+                    MyEntity entity = entityManager.find(MyEntity.class, itemId);
```
jpechane, line 171: There is no need for reindexing for inserts?

DDD-16.md
```diff
@@ -0,0 +1,143 @@
...
+quarkus.datasource.username=<your username>
+quarkus.datasource.password=<your password>
+quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/hibernate_orm_test
+quarkus.datasource.jdbc.max-size=16
+```
+
+With the cache-invalidation extension, the configuration should be reduced in the following:
+
+```properties
+quarkus.debezium.offset.storage=org.apache.kafka.connect.storage.MemoryOffsetBackingStore
```
mfvitale, line 60: This means that if the service that uses this extension is restarted the changes happening during the shutdown will not be replicated to the cache/index. Is what we want?

kmos: No, here we mentioned the options available for the extension and one of them is the offset storage. The configuration not available and defaulted are:
- snapshot
- in memory schema changes
- all the remaining configuration which are strings set to `invalidation`

DDD-16.md
```diff
@@ -0,0 +1,143 @@
...
+quarkus.datasource.username=<your username>
+quarkus.datasource.password=<your password>
+quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/hibernate_orm_test
+quarkus.datasource.jdbc.max-size=16
+```
+
+which are the datasource common configuration for agroal and the offset storage. The other configuration should be:
+
+- no snapshot
+- in memory Schema Changes
```
mfvitale, line 72: Could this be a problem with big databases, even if the table locks are minimal (mainly avoiding altering tables) , since every time the schema must be snapshotted?

DDD-16.md
```diff
@@ -0,0 +1,143 @@
+# Quarkus Debezium Cache/Search Invalidation
+
+Second‑level cache (L2C) and Search index can boost applications but becomes stale when database changes bypass the ORM without a way to notify applications of the changes. Debezium can capture those external changes from the DB transaction log and trigger near‑real‑time evictions of affected cache entries or a reindex. Debezium Extensions for Quarkus inside a Quarkus app allows registering a `@Capturing` method that can update the corresponding entity on CDC events. We propose a new Quarkus Extension that automatically creates the handlers for refreshing session-based data like hibernate L2C and search.
```
mfvitale, line 3: maybe instead of `eviction` we should just use update as a general concept since I think this extensions will not just remove things from the cache but add, update and remove them.

DDD-16.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Module Organization
+
+A developer should import the extension and the relative datasource for cache invalidation:
+
+```xml
+<dependency>
+    <groupId>io.debezium</groupId>
+    <artifactId>debezium-quarkus-hibernate-cache</artifactId>
```
mfvitale, line 12: Is this just a specific cache implementation example right? The extension is designed to support other cache provider like Redis, Infinispan etc?

DDD-16.md
```diff
@@ -0,0 +1,143 @@
...
+    <version>${version.debezium}</version>
+</dependency>
+```
+
+and in this way for reindex:
+
+```xml
+<dependency>
+    <groupId>io.debezium</groupId>
+    <artifactId>debezium-quarkus-hibernate-search-reindex</artifactId>
```
mfvitale, line 27: Cache and Search Index will be managed by the same extension or will be two different extensions?

DDD-16.md
kmos: if I understand correctly, for now we want the following extensions:

- hibernate 2LC
- hibernate search

in second instance a third extension that supports also the update of cache providers

DDD-16.md
kmos: After @jpechane comment, I thought that the best approach is to have two different extensions

DDD-16.md
kmos: Probably yes, is there any configuration that can solve this issue? Otherwise we can specify in the documentation this kind of limitation

mfvitale: I suppose that the user should be able to configure also the storage for the schema history

DDD-16.md
mfvitale: I agree

Naros: We probably should look at the dependency makeup to decide, but I suspect two. The reason is that if you want to only focus on L2C, there is no need to have any ties to Search in that setup.

DDD-16.md
Naros: So, for dev-services, we use the memory backing store unless overwritten by configuration. 
Is that right?

DDD-16.md
Naros: I suppose it would highly depend on what entities are marked as `@Cachable` or `@Indexed`. In practice, these are typically a much smaller subset of your overall entities, no?

kmos: maybe we can set as default the memory but the option configurable and available

DDD-16.md
mfvitale: So effectively we want to rely on the underlying hibernate 2LC that for now seems support only Infinispan and JCache (that have the different implementation). Right?

DDD-16.md
mfvitale: > maybe we can set as default the memory but the option configurable and available

Yes as for the offsets

DDD-16.md
Naros: I'd be very careful about updating the cache, mainly because entities can be loaded with dynamic fetch graphs and joins that may otherwise be marked as lazy. If we refresh the entity without that context, we'd likely evict it and only reload a subset of the original cached data.

DDD-16.md
Naros: Also, @kmos, does Hibernate expose in the EMF if the persistence unit has set the shared cache mode to `ALL`? This mode is something that can be set for the provider to cache all entities, avoiding the need to explicitly annotate the entities. I'm not sure if you've taken this into account.

DDD-16.md
kmos: > I suppose it would highly depend on what entities are marked as `@Cachable` or `@Indexed`. In practice, these are typically a much smaller subset of your overall entities, no?

I agree

DDD-16.md
kmos: Nope I can investigate and update the DDD!

DDD-16.md
kmos: I'll take note of it!

DDD-16.md
mfvitale: > I suppose it would highly depend on what entities are marked as `@Cachable` or `@Indexed`. In practice, these are typically a much smaller subset of your overall entities, no?

Are you referring to the table to capture? If yes I still think we should capture all schema otherwise if a new entity will be then marked as @Cachable or @Indexed we need to remove the schema history and redo the snapshot. Right?

DDD-16.md
kmos: updated

DDD-16.md
kmos: Yes

DDD-16.md
Naros: Yes. I do wonder though, does it make sense for users to have agency to say they only want the snapshot to focus on the `@Cachable` and `Indexed` annotated entities versus all entities? This way they can decide how the snapshot should work?

DDD-16.md
Naros: As long as we provide the user with the `Session` or `EntityManager` or expect them to inject that and use it to repopulate the cache based on the change event key details, then I think add/update/remove is fine.  

The user's implementation can decide what fetch query or join strategies to enable/disable at the CDC event time. But we should definitely delegate that to the user.

DDD-16.md
mfvitale: I think that we should start with some `opinionated` way (get schema of all table) with the idea of simply the works to the users. Then we could think to add some customization to let it be fine tuned on the user needs. WDYT?

DDD-16.md
kmos: I was thinking in the first iteration to use `evictEntityRegion` as eviction strategy to avoid issue with particular join strategies, but yes can be configurable (?)

DDD-16.md
beikov: You should probably also mention query cache invalidation by running something like this:

```java
SessionFactoryImplementor factory = emf.unwrap(SessionFactoryImplementor.class);
SharedSessionDelegatorBaseImpl lazySession = new SharedSessionDelegatorBaseImpl(null) {
	@Override
	public final SessionEventListenerManager getEventListenerManager() {
		return new SessionEventListenerManagerImpl( factory.getSessionFactoryOptions().buildSessionEventListeners() );
	}
	@Override
	public EventMonitor getEventMonitor() {
		return factory.getEventMonitor();
	}
};
factory.getCache().getTimestampsCache()
  .invalidate(new String[]{ "item_table" }, lazySession);
```

Though I wonder, couldn't the Debezium integration come with such an implementation out of the box?

Naros:
@kmos is this good to merge?

kmos:
> @kmos is this good to merge?

Yes :rocket: :rocket:

PR #26, opened 2026-03-23
GSoC 2026 : Milvus Source Connector for Debezium

yuanglili:
This is my GSoC 2026 proposal for Milvus Source Connector for Debezium
Zulip project discussion: https://debezium.zulipchat.com/#narrow/channel/573881-community-gsoc/topic/Yuang.20-.20Millvus.20Source.20Connector/with/581048390

Zulip introduction: [#community-gsoc > newcomers @ 💬](https://debezium.zulipchat.com/#narrow/channel/573881-community-gsoc/topic/newcomers/near/576880442)

Mentors
Giovanni Panice(@kmos ) and Chris Cranford (@Naros ) 

I will continue refining the proposal (formatting, wording, and structure), and I would really appreciate any feedback or suggestions.


gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+}
+```
+**Implementation notes:**
+The channel checkpoint stored in etcd is a Protobuf-encoded MQ seek position containing an embedded TSO timestamp. This is the anchor used to align snapshot and streaming 
+(see details in Technical Challenge 1).
+
+---
+##### Component 2: MilvusSchema
+Responsibility:
+- Implements Debezium's `DatabaseSchema<CollectionId>` interface to maintain the current Debezium schema state for all monitored collections.
```
Naros: Given that Milvus' collection schemas are more closely aligned with relational database schemas, do you think we could use `RelationalDatabaseSchema` or `HistorizedRelationalDatabaseSchema` rather than directly implementing the `DatabaseSchema` interface?

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+##### Approach:
+use the channel checkpoint stored in etcd as the anchor.
+Before the bulk read, read the MQ seek position from etcd and extract the embedded TSO as `guarantee_ts` for the SDK query(Milvus's Strong consistency level), Start streaming from exactly that MQ position after snapshot completes.
+
+**Milvus's Strong consistency level:**
+when a query is issued with a guarantee_ts,the QueryNode blocks until all writes before that timestamp are visible, giving the snapshot a well-defined point-in-time view.
+
+---
+
+### Key Components
```
Naros, line 134: There are a few components I would suggest thinking about and adding here:


**MilvusOffsetContext**
This is the transient offset state we keep synchronized, which the connector uses to create offset details in change events. 

What do you think is important here, and what do you think the offset structure might look like? This should consider the `guarantee_ts` you mentioned above, but also what is needed in order to maintain the position where the connector resumes from after a restart, to avoid duplicates?

**MilvusSourceInfoStructMaker**
This is the `Struct` creator that populates the `source` information block in change events. What data do you see relevant for events to track here? Does Milvus offer any metadata for each event that you believe might be useful for downstream consumers for debugging, analytics, etc.

**MilvusConnectorConfig**
You mention this below in one of the interfaces, and I think we should consider mapping out which configuration properties the user should provide. We can certainly change/add to this through the exercise, but do you have any early ideas?

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+| Week 3 (Jun 8–14) | - Implement `MilvusSchema` for in-memory schema tracking and Milvus-to-Debezium type conversion <br>- Define `MilvusMessageConsumer` abstraction <br>- Implement `KafkaMilvusMessageConsumer` <br>- Add offset serialization in `MilvusOffsetContext` | - Schema conversion works for all supported field types <br>- Kafka MQ consumption works in integration test <br>- Offset round-trip validated |
+| Week 4 (Jun 15–21) | - Implement `MilvusSnapshotChangeEventSource` <br>- Complete snapshot-to-streaming handoff using etcd checkpoint and `guarantee_ts` <br>- Emit snapshot events through Debezium's event pipeline | - Snapshot works against live Milvus <br>- Streaming start position recorded correctly |
+| Week 5 (Jun 22–28) | - Implement `MilvusStreamingChangeEventSource` <br>- Add `MilvusChangeEventSourceFactory` <br>- Wire connector into Debezium's source pipeline end to end | - First end-to-end prototype working <br>- Milvus change → Debezium event flow validated |
```
Naros: The streaming component of any CDC connector is the most important. Given that your timeline is already full, I would suggest focusing on the snapshot-to-streaming handoff from etcd, then jumping right into streaming.  We can circle back and add capturing the historical data toward the end. wdyt?

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+I do not expect any major external commitments during the GSoC coding period beyond my regular coursework. I will prioritize this project accordingly and will communicate early if any scheduling conflicts arise.
+
+## Appendix
+
+The following edge cases are handled within `MilvusStreamingChangeEventSource` (see Component 7) and `MilvusSnapshotChangeEventSource` (see Component 6):
+
+- **Drop collection with buffered DML** (Case 5): If a `DropCollectionMsg` is received while DML events for that collection are still buffered, all buffered DML events are discarded, `MilvusSchema.removeSchema()` is called, and a drop schema change event is emitted. 
+
+- **Timetick interruption** (Case 6): If no `TimeTickMsg` is received within a configurable timeout, buffered events are force-flushed. If the stall persists beyond a second threshold, the connector stops and waits for manual intervention.
```
Naros: Could this be handled more gracefully by logging a warning rather than stopping?

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+- **MQ offset expired** (Case 7): If the stored MQ checkpoint position is no longer available on the broker, the connector stops with a descriptive error and waits for manual intervention.
```
Naros: See `snapshot.mode` and `Snaphotter` API in the main repository. This behavior would be determined by the chosen snapshot mode. In some modes, if the MQ offset has expired, you start treating the connector as if its a brand new deployment.

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Other commitments
+
+I do not expect any major external commitments during the GSoC coding period beyond my regular coursework. I will prioritize this project accordingly and will communicate early if any scheduling conflicts arise.
+
+## Appendix
+
+The following edge cases are handled within `MilvusStreamingChangeEventSource` (see Component 7) and `MilvusSnapshotChangeEventSource` (see Component 6):
+
+- **Drop collection with buffered DML** (Case 5): If a `DropCollectionMsg` is received while DML events for that collection are still buffered, all buffered DML events are discarded, `MilvusSchema.removeSchema()` is called, and a drop schema change event is emitted.
```
Naros: The Milvus CDC approach works this way because its goal is to replicate data. In their use case, they see it as wasted IO to replicate the DML events when the collection is about to cease to exist.

But when you're thinking about this beyond replication, there are reasons to emit the DML events rather than discard them. We need to weigh the complexities of how to handle this, but I think that should be our ideal goal and falling back to discard if and only if that's the only alternative. This keeps the behavior of the Milvus connector aligned with all the other Debezium source connectors. wdyt?

Naros (commented):
Hi @yuanglili this was a great read. I've left a few comments below, if you could take a look and update the proposal as needed. If you have any follow-up questions, please feel free to ping.

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Project Information
+
+### Abstract
+
+Milvus is a cloud-native vector database for high-performance similarity search. Debezium currently provides a Milvus sink connector, but not a source connector, so changes within Milvus cannot yet be captured for replication, auditing, or downstream processing.
+
+This project proposes a Debezium source connector for Milvus 2.5, covering initial snapshot, streaming CDC, schema handling, offset management, and restart recovery. Kafka is the primary target, with Pulsar support considered as a stretch goal.
+
+Milvus 2.5 is selected because its Kafka/Pulsar-based architecture is comparatively stable and well understood. Milvus 2.6 introduces Woodpecker and a new StreamingService layer, which significantly changes the consumption model and still appears to be evolving, so support for 2.6 is deferred as future work.
```
kmos, line 37: Can you provide more details about the business value of a Milvus Connector? What kinds of use cases can it solve?

gsoc/yuangLi/YuangLiProposal.md
```diff
@@ -0,0 +1,364 @@
...
+  ) throws InterruptedException;
+}
+```
+
+---
+
+##### Testing Strategy
+
+- Unit tests: cover isolated logic: deserialization, schema conversion, and offset serialization
+- Integration tests: use Testcontainers to validate etcd reads, MQ consumption, snapshot, and streaming against real infrastructure
```
kmos, line 461: Is there already a Testcontainer for Milvus?

kmos (commented):
Thank you for the contributions you’ve made to Debezium. We really appreciate the time you’ve dedicated to the project!

gsoc/yuangLi/YuangLiProposal.md
yuanglili: @Naros Based on my current understanding, I think DatabaseSchema is a better fit

For `RelationalDatabaseSchema`, my main hesitation was around the schema-building pipeline through `TableSchemaBuilder`, `ValueConverterProvider`, and `jdbcType().` Milvus types such as `FloatVector` and `SparseFloatVector` do not seem to map to `java.sql.Types`, so I wasn’t sure how to integrate them cleanly without forcing them into an abstraction that is fundamentally relational.

For `HistorizedRelationalDatabaseSchema`, my understanding from the source is that recovery relies on replaying `SchemaHistory` through a `DdlParser`. In the Milvus case: the schema is recoverable directly from `etcd`, and the `CreateCollectionRequest` message in message queue already carries the full `CollectionSchema` inline. Because of that, maintaining **a separate DDL history layer** seemed unnecessary, if etcd serves as the authoritative schema source.

The tradeoff with current approach is that it does not reuse the built-in schema `registry` that `RelationalDatabaseSchema` provides. In `MilvusSchema`, I’m planning to handle that with a `ConcurrentHashMap<CollectionId, DataCollectionSchema>` instead. That felt acceptable because, from what I could tell in Milvus 2.5, schema evolution is relatively limited compared to relational systems, and does not seem to follow a broad ALTER TABLE-style model.

This is my current thinking, but this is just my current understanding. so please let me know if I’m missing something important.

Naros: > For `RelationalDatabaseSchema`, my main hesitation was around the schema-building pipeline through `TableSchemaBuilder`, `ValueConverterProvider`, and `jdbcType()`. Milvus types such as `FloatVector` and `SparseFloatVector` do not seem to map to `java.sql.Types`, so I wasn’t sure how to integrate them cleanly without forcing them into an abstraction that is fundamentally relational.

If you take a look at the `BinlogValueConverters` class for an example, you'll see that in many cases, we don't even rely on the `jdbcType()` but rather the `typeName()` for the mappings. In other databases, such as PostgreSQL, there is a `TypeRegistry` that maps ObjectIDs to specific database types, such as PG's Double, Float, and Sparse vector types. So using some sort of `ValueConverterProvider` is vital. 

That's not to say we need to use the `RelationalDatabaseSchema`, but we should strive to reuse as much as possible from the common connector framework.

> For `HistorizedRelationalDatabaseSchema`, my understanding from the source is that recovery relies on replaying `SchemaHistory` through a `DdlParser`. 

I would say that if there is a reason to just ingest the DDL events without using a DDL parser, that's doable. There is already an internal option for this that we could use if needed.

> In the Milvus case: the schema is recoverable directly from etcd, and the `CreateCollectionRequest` message in message queue already carries the full `CollectionSchema` inline. Because of that, maintaining a separate DDL history layer seemed unnecessary, if etcd serves as the authoritative schema source.

PostgreSQL operates the exact same way, and it's one reason it does not use the historized impl right now. There have been multiple requests for PostgreSQL to at least emit schema change events so downstream consumers can ingest them, and one could argue the same might apply here. It's not something we need to do as part of the initial GSoC work, but I would like it to be something we consider in the design, so we can revisit it and provide a solution if there is user demand.

> The tradeoff with current approach is that it does not reuse the built-in schema `registry` that `RelationalDatabaseSchema` provides. In `MilvusSchema`, I’m planning to handle that with a `ConcurrentHashMap<CollectionId, DataCollectionSchema>` instead. That felt acceptable because, from what I could tell in Milvus 2.5, schema evolution is relatively limited compared to relational systems, and does not seem to follow a broad ALTER TABLE-style model.

I assume your plan is to mimic the behavior of `MongoDbSchema` 's `collections` map?

yuanglili: > If you take a look at the `BinlogValueConverters` class for an example, you'll see that in many cases, we don't even rely on the `jdbcType()` but rather the `typeName()` for the mappings. In other databases, such as PostgreSQL, there is a `TypeRegistry` that maps ObjectIDs to specific database types, such as PG's Double, Float, and Sparse vector types. So using some sort of `ValueConverterProvider` is vital.
> 
Thank you! I think I hadn’t looked at that part of the code carefully enough before.

The dispatch doesn’t need to go through `jdbcType()` 
- So a `MilvusValueConverter` implementing `ValueConverterProvider` can be used.
- I also found that `TableSchemaBuilder` can be reused directly by mapping `FieldSchema` to `Column` using `name`, `typeName`, `optional`, and `position`, with `jdbcType` set to `Types.OTHER` as a placeholder.

- My remaining concern is that `RelationalDatabaseSchema` is fixed to `DatabaseSchema<TableId>`, so using it with `CollectionId` would mean introducing an artificial mapping that does not feel very justified.

>  For `HistorizedRelationalDatabaseSchema`, my understanding from the source is that recovery relies on replaying `SchemaHistory` through a `DdlParser`.
 PostgreSQL operates the exact same way, and it's one reason it does not use the historized impl right now. There have been multiple requests for PostgreSQL to at least emit schema change events so downstream consumers can ingest them, and one could argue the same might apply here. It's not something we need to do as part of the initial GSoC work, but I would like it to be something we consider in the design, so we can revisit it and provide a solution if there is user demand.
> 
That sounds interesting. I can update `applySchemaChange()` to accept an `EventDispatcher` parameter and add the implementation later if needed.

> I assume your plan is to mimic the behavior of `MongoDbSchema` 's `collections` map?
> 
Yes, my origin plan is to mimic MongoDbSchema's collections map.

My current idea is to keep `MilvusSchema` as `DatabaseSchema<CollectionId>`, while reusing `TableSchemaBuilder`, introducing a `MilvusValueConverter`, and allowing `applySchemaChange()` to accept an `EventDispatcher`.Are there any additional references I should look at, or anything important I might be missing or not considering?

gsoc/yuangLi/YuangLiProposal.md
yuanglili: > **MilvusOffsetContext**
>
MilvusOffsetContext needs to serve two purposes: 
- storing the snapshot anchor point
- storing the MQ position so `MilvusStreamingChangeEventSource` can seek to the correct position on restart

So the offset structure will be:
- `vchannel`
- `guarantee_ts`
- `snapshot_completed`
- MQ position :
    Kafka: `mq_topic`, `mq_partition` (int), `mq_offset` (long)
    Pulsar: `pulsar_ledger_id` (long), `pulsar_entry_id` (long)

My current idea is that the offset is tracked per `vchannel`, since each vchannel has its own streaming position.

On restart, `Loader` reads these fields back and recovers as follows:
- if `snapshot_completed` is `false`, the connector re-executes the full snapshot with a fresh `guarantee_ts` read from etcd; 
- if `true`, `MilvusStreamingChangeEventSource` seeks directly to the stored MQ position.

yuanglili: > MilvusSourceInfoStructMaker

I think the fields in the source block can be grouped into three categories:

1. Source identity (where the data comes from)
`collection`, `partition`, `vchannel`
-  for routing, filtering, and shard-level debugging.
2. Position and ordering (how to locate and order events)
`tso`, `segment_id`
- `tso` provides ordering and latency insight,  
- `segment_id` helps with replay and tracing.

3. Event metadata (additional context for debugging and analytics)
`collection_id`, `partition_id`, `msg_id`,`num_rows`, `primary_keys`（for delete events）
- `collection_id`, `partition_id`,  help distinguish resources after drop/recreate, 
- `msg_id` — unique per message, for deduplication and debugging in at-least-once scenarios.
-  `num_rows` / `primary_keys` reflect batch-level information.

yuanglili: > MilvusConnectorConfig

For `MilvusConnectorConfig`, I have some ideas on the required properties but I think it would be cleaner to update the proposal directly with a structured breakdown 
I will group them by connection, filters, snapshot, and advanced, using ts MongoDbConnectorConfig and CassandraConnectorConfig as a reference. I'll push an update tomorrow.

gsoc/yuangLi/YuangLiProposal.md
yuanglili: I'll incorporate `snapshot.mode` into `MilvusConnectorConfig` and tie the expired offset recovery behavior to the chosen mode

gsoc/yuangLi/YuangLiProposal.md
yuanglili: Makes sense. It's a really good point. I’ll revisit my design and make sure each part aligns with this perspective as much as possible.

gsoc/yuangLi/YuangLiProposal.md
Naros: > My remaining concern is that RelationalDatabaseSchema is fixed to DatabaseSchema<TableId>, so using it with CollectionId would mean introducing an artificial mapping that does not feel very justified.

Can you elaborate on what you mean? What artificial mapping are you referring?

The last thing I'll say is that it feels a bit foreign to me to think about using only specific parts of the `io.debezium.relational` package, but this is something we can work through during coding. It's not super important at this junction as it's merely an implementation detail.

gsoc/yuangLi/YuangLiProposal.md
yuanglili: @Naros @kmos Yes, MilvusContainer is available in the Testcontainers library, Debezium is already using it in its integration tests 
[MilvusSchemaTest](https://github.com/debezium/debezium-server/blob/9e0f6b79907afc16d184e425134b6f63ecbe47a3/debezium-server-milvus/src/test/java/io/debezium/server/milvus/MilvusSchemaTest.java#L18)
[MilvusTestResourceLifecycleManager](https://github.com/debezium/debezium-server/blob/9e0f6b79907afc16d184e425134b6f63ecbe47a3/debezium-server-milvus/src/test/java/io/debezium/server/milvus/MilvusTestResourceLifecycleManager.java#L14)

gsoc/yuangLi/YuangLiProposal.md
Naros: > My current idea is that the offset is tracked per vchannel, since each vchannel has its own streaming position.

Let's say Milvus has 2 shards with 50 collections, which means you need to track 100 independent vchannel data sets, whereas the number of pchannels is often significantly smaller, ~16 on average. You are correct to assume you need to track the vchannel state. I'm just curious whether vchannel at the top is the right call, since you're always consuming from different pchannels that can represent N vchannels.

We don't need to solve it right now, but we need to consider complex deployments and Milvus's scalability and how they impact the data flow.

> * `num_rows` / `primary_keys` reflect batch-level information.

This is an interesting point with batch-level operations. If this refers specifically to deletes, would it make more sense to emit one delete event per primary key tuple rather than a single CDC event representing a batch? This at least aligns more closely with what we do for other databases, and fits very nicely into Kafka topic compaction capabilities & consumer sink pipeline expectations. wdyt?

gsoc/yuangLi/YuangLiProposal.md
yuanglili: > My remaining concern is that RelationalDatabaseSchema is fixed to DatabaseSchema, so using it with CollectionId would mean introducing an artificial mapping that does not feel very justified.

What I meant is there’s a mismatch between TableId (catalog/schema/table) and CollectionId (database/collection). Using `RelationalDatabaseSchema` would require a placeholder, which would carry through things like topic naming and filtering and lead to awkward names for downstream consumers.


> The last thing I'll say is that it feels a bit foreign to me to think about using only specific parts of the `io.debezium.relational` package,

I agree with you. Taking another look, the mismatch between TableId and CollectionId may not be as complex as I initially thought.  I can prototype the mapping before implementation,if it turns out to be straightforward, we can go with `RelationalDatabaseSchema` directly, otherwise we can  fall back to implementing `DatabaseSchema`

Naros: > What I meant is there’s a mismatch between TableId (catalog/schema/table) and CollectionId (database/collection). Using RelationalDatabaseSchema would require a placeholder, which would carry through things like topic naming and filtering and lead to awkward names for downstream consumers.

Take a look at MySQL/MariaDB and PostgreSQL. The former leaves the `schema` empty, and the latter leaves the `database` field empty. So that pattern is already well established. For Milvus, I suspect you'd likely follow a similar pattern to MySQL and MariaDB if you wanted to. 

You're certainly welcome to avoid using it if you think it makes sense, but I didn't want you to shy away from its usage if that is the only reason.

> if it turns out to be straightforward, we can go with RelationalDatabaseSchema directly, otherwise we can fall back to implementing DatabaseSchema

Perfect, that's a great idea. I can't wait to see your pros/cons for each :tada:

yuanglili:
@Naros @kmos I pushed an update based on your feedback. I really appreciate the detailed review, it helped a lot. Could you please take another look when you get a chance?

Main changes:
- `MilvusSchema` updated from `DatabaseSchema<CollectionId>` to `RelationalDatabaseSchema` with `MilvusValueConverter`; 
-  added a pros/cons analysis between `DatabaseSchema<CollectionId>` and `RelationalDatabaseSchema`under Technical Challenges （the framework integration prototype may not make it before the proposal deadline, but I'll have it ready before coding starts）
- Added `MilvusOffsetContext`, `MilvusPartition`, `MilvusSourceInfoStructMaker`, and `MilvusConnectorConfig` as new components
- Updated edge case handling for drop+DML, timetick stall, and expired MQ offset
- Updated roadmap to prioritize streaming over snapshot per your suggestion
- Expanded abstract to include business value and user-facing use cases

Happy to make more updates if anything looks off!

kmos (approved):
Thanks, @yuanglili, for your contributions and all the work you’ve done! We truly appreciate your efforts throughout this phase of GSoC. Once you’ve addressed @Naros’s comments, you can officially submit your proposal to GSoC.

Naros (approved):

PR #27, opened 2026-03-24
GSOC 2026: SQLite Connector

siddhantcvdi:
This is my GSoC 2026 proposal for a SQLite Connector for Debezium.


kmos:
@siddhantcvdi any update?

siddhantcvdi:
@kmos I’ve completed the draft. Apologies for the delay, I was dealing with some health issues.

gsoc/2026/siddhantcvdi/proposal.md
```diff
@@ -0,0 +1,688 @@
...
+To know the schema of a table, the connector should read this SQL string and parse it, extracting column names, declared types, and constraints.
+
+### The Rowid Alias
+When a column is declared `INTEGER PRIMARY KEY`, SQLite makes it an alias for the table's internal rowid and omits it from the record body entirely. If the connector does not account for this, it reads NULL for that column. We need to flag such columns during parsing so the WAL page decoder knows to substitute the rowid value rather than reading from the record body.
+
+### Root Page Mapping
+The `rootpage` value tells the connector which page number in the .db file is the root of that table's B-tree. The connector knows the page number that changed but not which table it belongs to. So we need to maintain a root page to table name map to identify which table a WAL frame is modifying.
+
+### Schema Change Detection
+SQLite increments PRAGMA schema_version on every DDL operation. The connector can poll this value each streaming cycle. When it changes, SqliteDatabaseSchema re-reads sqlite_master, rebuilds the column map, and emits a schema change event.
```
kmos, line 174: the `sqlite_master` changes aren't reflected in the WAL ? Can you read the delta from the WAL and identify a schema change?

gsoc/2026/siddhantcvdi/proposal.md
```diff
@@ -0,0 +1,688 @@
...
+- Unit tests for each WAL parsing class with known binary inputs and expected outputs
+- Integration tests covering all major CDC scenarios end-to-end
+- README with setup instructions, config reference, and deployment guide
+- Documentation of known limitations
+
+## Known Limitations
+
+The following limitations are inherent to SQLite's architecture or to the scope of this project. They are documented explicitly rather than treated as bugs to be fixed later.
+ 
+| Limitation | Reason |
```
kmos, line 544: Due to these limitations, could you mention other solutions for achieving CDC in SQLite? Also, could you investigate how other systems (such as Turso) implement CDC?

kmos (commented):
I just wanted to thank you for all the contributions you’ve made to Debezium. We really appreciate the time and effort you’ve put into the project.

My last suggestion before submit your application is to highlight drawbacks and alternative of your solution.

Thanks again for all your hard work!

vsantonastaso (commented):
Hi @siddhantcvdi,
First of all, thanks a lot for your time and engagement. I have a couple of questions and concerns that came to mind while reading through the proposal:

Transactions and event grouping
SQLite transactions can atomically modify multiple tables. Do events produced from a single transaction preserve any ordering? Are BEGIN / COMMIT events emitted? More generally, how would a consumer know which events belong to the same transaction?
Debezium already supports transaction metadata, do you see this as something planned for this connector, or is it considered out of scope?

WITHOUT ROWID: in or out?
The timeline (Week 9) mentions implementing WITHOUT ROWID support, but the Known Limitations section lists it as deferred Phase 2 work. This is a bit unclear. How do you plan to handle it?
This matters because the current diff-by-rowid approach would need a fundamental redesign. I’d suggest either committing to it with a dedicated design section, or removing it from the timeline for now.

Page cache growth
The page cache (Map<pageNumber, lastFrameData>) stores full pages and grows until a checkpoint. With a high-write workload touching thousands of different pages between checkpoints, this could become problematic.
What happens if the cache grows too large? Have you considered a size limit, an eviction policy, or forcing a checkpoint from the connector side?

Thanks again.

siddhantcvdi:
Hi @vsantonastaso ,

Thank you for the detailed review!

1. WAL frames are processed sequentially within a committed block and the ordering of events within a transaction is preserved. In my opinion it's better if I support transaction metadata. This allows consumers to group events that occurred within the same commit. 
Suppose we had a transaction - 
```sql
BEGIN; 
UPDATE table1... 
UPDATE table2....
COMMIT;
```
The connector will detect the Commit Marker in the WAL and emit a begin event, two DATA events sharing the same txId, and an end event.

2. Earlier I was a bit confused about that case but Ilooked at the spec. The record format inside cells is identical. the only differences are skipping the rowid varint, using PK columns as the diff key, and a different overflow threshold formula. It's roughly 1-2 days of additional work and I'll include it in Week 9 alongside the overflow page implementation. 

3. The checkpoint mechanism bounds the cache by design. The periodic controlled checkpoint clears the page cache entirely. We can have a default interval of 1000 frames with 4096 bytes per page, the worst case before a cache clear is ~4MB. Additionally we can have a checkpoint interval config. Users with high write throughput can tune it  lower. As an additional safety net, I will add a configurable max cache size that tracks total cache size on every write and forces an early controlled checkpoint immediately if the limit is exceeded,

I'll mention these things in the proposal once approved. Thanks!!!

gsoc/2026/siddhantcvdi/proposal.md
siddhantcvdi: @kmos Schema changes modify sqlite_master, and those page updates are written to the WAL as well. But the WAL only contains raw page level diffs, not  information about what actually changed. It is technically possible to detect that a sqlite_master page was modified. But reconstructing a schema change from WAL alone would require decoding the page, extracting the SQL definition, and diffing it against the previous version. That would be quite complex, especially for cases like column reordering, constraints, or partial changes.

Because of that, I’m treating WAL purely as a source of row-level data changes, and relying on PRAGMA schema_version as a lightweight and reliable signal that a DDL operation

gsoc/2026/siddhantcvdi/proposal.md
siddhantcvdi: @kmos I'll add this section in Proposal

1. Trigger-based CDC
Tables can be instrumented with INSERT/UPDATE/DELETE triggers that write changes into a separate log table. This is simple  but it requires modifying the user schema and adds overhead on every write. It also doesn’t work well for existing databases where users would not want schema changes.

2. Application-level hooks
SQLite exposes hooks that notify about row changes. However, these are connection-local and only work if the application cooperates. They cannot observe changes made by other processes.

3. Polling-based approaches
Periodically querying tables using timestamps or rowid ranges can be used but it is inefficient/.

Regarding systems like Turso: from what I’ve been able to understand, they don’t rely on raw WAL decoding for CDC. Turso is built on libSQL, which extends SQLite with a replication layer. They capture changes at a higher level within the engine itself and stream them as logical updates instead of reconstructing them from page diffs. Because they control the database engine, they can expose structured change information directly.

In contrast, a Debezium connector has to work with stock SQLite , without modifying the engine or requiring application changes. That’s why WAL parsing in my opinion is the most suitable approach.

siddhantcvdi:
> I just wanted to thank you for all the contributions you’ve made to Debezium. We really appreciate the time and effort you’ve put into the project.
> 
> My last suggestion before submit your application is to highlight drawbacks and alternative of your solution.
> 
> Thanks again for all your hard work!

Thanks a lot @kmos for the review. I'll add this section in my proposal. I wanted to ask, if the proposal length is fine. I did not pay a lot of attention to the length of proposal while writing as i wanted to go into a good amount of technical depth.

kmos:
> > I just wanted to thank you for all the contributions you’ve made to Debezium. We really appreciate the time and effort you’ve put into the project.
> 
> > 
> 
> > My last suggestion before submit your application is to highlight drawbacks and alternative of your solution.
> 
> > 
> 
> > Thanks again for all your hard work!
> 
> 
> 
> Thanks a lot @kmos for the review. I'll add this section in my proposal. I wanted to ask, if the proposal length is fine. I did not pay a lot of attention to the length of proposal while writing as i wanted to go into a good amount of technical depth. 

@siddhantcvdi , in my opinion you shouldn't focus on the length. My suggestion was about the content. We don't know in this phase if everything would work but we can have in mind drawbacks.

siddhantcvdi:
> @siddhantcvdi , in my opinion you shouldn't focus on the length. My suggestion was about the content. We don't know in this phase if everything would work but we can have in mind drawbacks.

Yuppp..I get it, I thought my proposal was too long😅

vsantonastaso:
> Hi @vsantonastaso ,
> 
> Thank you for the detailed review!
> 
> 1. WAL frames are processed sequentially within a committed block and the ordering of events within a transaction is preserved. In my opinion it's better if I support transaction metadata. This allows consumers to group events that occurred within the same commit.
>    Suppose we had a transaction -
> 
> ```sql
> BEGIN; 
> UPDATE table1... 
> UPDATE table2....
> COMMIT;
> ```
> 
> The connector will detect the Commit Marker in the WAL and emit a begin event, two DATA events sharing the same txId, and an end event.
> 
> 2. Earlier I was a bit confused about that case but Ilooked at the spec. The record format inside cells is identical. the only differences are skipping the rowid varint, using PK columns as the diff key, and a different overflow threshold formula. It's roughly 1-2 days of additional work and I'll include it in Week 9 alongside the overflow page implementation.
> 3. The checkpoint mechanism bounds the cache by design. The periodic controlled checkpoint clears the page cache entirely. We can have a default interval of 1000 frames with 4096 bytes per page, the worst case before a cache clear is ~4MB. Additionally we can have a checkpoint interval config. Users with high write throughput can tune it  lower. As an additional safety net, I will add a configurable max cache size that tracks total cache size on every write and forces an early controlled checkpoint immediately if the limit is exceeded,
> 
> I'll mention these things in the proposal once approved. Thanks!!!

It's not necessary to mention every concern at this step in your proposal. 
In my opinion the important thing is to keep this concern in mind, since it can also help during implementation. 
The idea is to iterate while keeping things open to the changes so we can address them in the future. 
BTW Thanks for the feedback.

vsantonastaso (approved):

kmos (approved):

PR #30, opened 2026-03-25
GSoC 2026: Proposal for Host-Based Pipeline Deployment — Divyanshu Kumar

d1vyanshu-kumar:
## GSoC 2026 Proposal: Host-Based Pipeline Deployment for the Debezium Platform

Hi! I'm Divyanshu Kumar (@d1vyanshu-kumar), and this is my proposal for the **Host-Based Pipeline Deployment** project under GSoC 2026.

### What this proposal covers

The Debezium Platform currently supports deploying pipelines on Kubernetes via the Debezium Operator. This proposal extends that to **bare-metal servers, VMs, and cloud instances** - environments where Kubernetes isn't available -using SSH for remote access and Docker for container-based deployment.

[Google Doc](https://docs.google.com/document/d/1ql_l3J9bLcH8ecCzNXNWUjLP5Ou9ELYR4L6Lfqp02Qs/edit?usp=sharing)

Looking forward to feedback from the mentors! 🙏
@mfvitale 
@kmos


gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+| [Issue #1242](https://github.com/debezium/dbz/issues/1242) | When CDC is enabled at the database level but no tables are tracked yet, Debezium fails to start. Instead of recognizing this valid state, it logs a misleading error claiming the user doesn't have CDC access, even when they do. | [https://github.com/debezium/debezium/pull/7132](https://github.com/debezium/debezium/pull/7132) | Merged |
+| [Issue #74](https://github.com/debezium/dbz/issues/74) | Added support for plugging in a custom `EnvelopeSchemaFactory` to define the Kafka Connect schema of the Debezium change event envelope. | [https://github.com/debezium/debezium/pull/7214](https://github.com/debezium/debezium/pull/7214) | Open |
+
+Please check out my other PR (Open Source) from here:
+[https://docs.google.com/document/d/1mP5n6P_U8J4taswqQXOP4xJ-wegROmPZDi5m2TqyFyw/edit?usp=sharing](https://docs.google.com/document/d/1mP5n6P_U8J4taswqQXOP4xJ-wegROmPZDi5m2TqyFyw/edit?usp=sharing)
+
+**Feedback from my last GSoC'25:**
+
+- [https://summerofcode.withgoogle.com/evaluations/mtz59fQC](https://summerofcode.withgoogle.com/evaluations/mtz59fQC)
```
mfvitale, line 812: @d1vyanshu-kumar the access seems restricted.

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+**Credential Security:** SSH credentials must never be stored in plain text. I'll integrate with the platform's existing secrets management (which uses Kubernetes Secrets for the Kubernetes path). For host-based deployments, I'll store SSH key references - not raw keys - in the database, and the actual keys are loaded from a configurable secrets backend (environment variable, filesystem path, or a secrets manager like HashiCorp Vault via a pluggable `SecretProvider` interface). This way the database never holds sensitive material.
+
+**Connection Pooling:** Opening a new SSH connection for every operation is expensive. I'll implement a simple `SshConnectionPool` that maintains idle authenticated sessions per host and reuses them for repeated commands (like status checks that run on a schedule).
+
+---
+
+##### Component 3: The Host Provisioning Service (Preparing the Environment)
+
+**What this is:** When a user registers a new host as a deployment target, the platform needs to validate that host and prepare it before any pipeline can be deployed there. This is the automated provisioning requirement.
```
mfvitale: > When a user registers a new host

Could you elaborate more on this? How the user can define the target hosts?

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+This requirement ensures that all the configuration for a pipeline - the source database connection details, the destination config, the Debezium Server version, the connector settings - is managed centrally through the Debezium Platform, not scattered across individual servers.
+
+This matters because if a user has Debezium pipelines running on five different hosts, they shouldn't have to SSH into each one separately to update a configuration. They should change it once in the Platform and the Platform pushes the update everywhere.
+
+My implementation needs to store pipeline configurations in the Platform's existing data store, version them, and ensure they're correctly applied whenever a deployment is created or updated on any host.
+
+---
+
+#### Requirement 5 (Optional): REST API for Lifecycle Management
```
mfvitale, line 137: Well this part is more a separate component to be deployed in the host machine so that the `conductor` just needs an initial SSH session to setup all the things and then the DS lifecycle is managed vie this REST API.

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+A deployment system that touches live infrastructure - SSH connections, Docker daemons, remote filesystems - is exactly the kind of system where an untested change can silently break a production pipeline. For that reason, testing is not an afterthought in this project. It is a first-class deliverable, designed alongside the implementation, not bolted on at the end.
+
+My testing strategy is structured into three layers: unit tests that run instantly with no external dependencies, integration tests that validate real SSH and Docker behavior using isolated containers, and API-level tests that verify the REST endpoints behave correctly from end to end. Each layer has a distinct job, and together they create a safety net that covers every critical path in the system.
+
+**Coverage Goals and Tooling:**
+
+I'll introduce [JaCoCo](https://www.jacoco.org/jacoco/) for measuring code coverage, integrated with Maven so that coverage reports are generated automatically on every build. The existing project does not currently use JaCoCo, so this would be a new addition to the build - one that benefits the entire project, not just the host-deployment module.
```
mfvitale: Do you think that instead of jacoco you can work on `REST API for Lifecycle Management`?

mfvitale:
@d1vyanshu-kumar First of all thanks you so much for you hard work contributing to Debezium, we appreciate it a lot. 

Regarding the proposal I left some clarification questions but overall it LGMT.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: Yes, you’re right. I shared my GSoC portal profile link, which is restricted to others. I’ll include a screenshot along with last year’s public contributor link instead.

https://summerofcode.withgoogle.com/archive/2025/projects/mW5rZglq

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: Thanks for the question! The user would define target hosts through the platform in a similar way they currently define Sources and Destinations, via the REST API (and eventually the Stage UI).

Specifically, I plan to add a HostTargetResource with a POST /api/hosts endpoint. 
The user provides:
- name (a human-readable label like "production-server-1")
- hostname (IP or DNS of the remote machine)
- port (default 22)
- username (SSH user)
- authentication type (SSH_KEY or PASSWORD)
- credential reference (pointing to a Vault entry containing the private key or password)

Once registered, the platform automatically triggers provisioning, SSHing into the host, validating it (Docker installed, daemon running, ports available, etc.), and reporting back the result. 

When creating a pipeline, the user selects a registered host as the deployment target. This association determines whether the pipeline is deployed via the Kubernetes path (OperatorPipelineController) or the host-based path (HostPipelineController). 

For the initial implementation, this would be API-driven. Stage UI integration (a "Register Host" form and a target selector in the pipeline creation wizard) could follow as a natural next step.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: That makes a lot of sense, thank you for the clarification! So the idea is that during provisioning, the Conductor uses SSH to set up the remote host, install Docker, verify the environment, and then deploy this lightweight service component on the host itself. Once the agent is running, the Conductor switches from SSH to HTTP for all subsequent lifecycle operations. So the flow would be:

1. Provisioning (SSH): Conductor connects, validates the environment, and installs Docker if needed, and deploys the agent as a service (e.g., via systemd)
2. Lifecycle management (HTTP): Conductor calls the agent's REST API, POST /deploy, GET /status, POST /stop, etc., and the agent handles 

This way, the Conductor only needs SSH for the initial setup. All ongoing lifecycle management goes through the agent's API, which is cleaner, faster (no SSH overhead per operation), and gives the agent the ability to manage things locally, like watching for container crashes and restarting automatically.
 
I'll design the agent as a lightweight Quarkus application that gets bundled as part of the provisioning artifacts and started as a systemd service on the target host. The HostPipelineController on the Conductor side would then be an HTTP client calling this agent's API, rather than executing Docker commands over SSH directly.

I'll update the proposal to reflect this architecture. Thanks for pointing this out, it's a much better separation of concerns.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: Absolutely, that's a much better use of the time. I'll drop JaCoCo from the deliverables and focus that effort on building the service component for host-based lifecycle management instead.

Based on your earlier feedback, I'll design it as a lightweight service that gets deployed on the target host during provisioning and exposes REST APIs for deploy, update, stop, remove, status, and logs. The Conductor would use SSH only for the initial setup (provisioning + agent installation), and then switch to HTTP calls to this agent for all ongoing lifecycle operations.

I'll update the timeline to reflect this, the weeks previously allocated to Coverage tooling will go toward agent design, implementation, and integration testing instead.

d1vyanshu-kumar:
> @d1vyanshu-kumar First of all thanks you so much for you hard work contributing to Debezium, we appreciate it a lot.
> 
> Regarding the proposal I left some clarification questions but overall it LGMT.

@mfvitale  Thanks a lot for your feedback! I’ll update my proposal before submitting it on the GSoC portal.

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+I want to contribute to Debezium well beyond this summer. This proposal is the start of that.
+
+---
+
+## 3. My Background
+
+Hi, I'm Divyanshu Kumar - a third-year Computer Science student from India, majoring in AI & ML at Starex University.
+
+I got into open source in my very first year of college, and it stuck. I've contributed to projects across a pretty wide range, including OpenWISP, Sugarizer, Jenkins, Headlamp (a Kubernetes dashboard), MIT App Inventor, and Debezium among others. Each contribution taught me something different - not just technically, but about how real software teams collaborate and ship things.
```
kmos, line 61: Please share other contributions that you have made

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+    String stdout = readStream(channel.getInvertedOut());
+    String stderr = readStream(channel.getInvertedErr());
+    int exitCode = channel.waitFor(EnumSet.of(ClientChannelEvent.CLOSED), EXEC_TIMEOUT);
+
+    return new CommandResult(stdout, stderr, exitCode);
+}
+```
+
+**Credential Security:** SSH credentials must never be stored in plain text. I'll integrate with the platform's existing secrets management (which uses Kubernetes Secrets for the Kubernetes path). For host-based deployments, I'll store SSH key references - not raw keys - in the database, and the actual keys are loaded from a configurable secrets backend (environment variable, filesystem path, or a secrets manager like HashiCorp Vault via a pluggable `SecretProvider` interface). This way the database never holds sensitive material.
```
kmos: keyring can be a simple alternative

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
+
+Fourth, that the system fails gracefully - a refused SSH connection, a missing Docker installation, a container that crashes immediately after starting - all of these produce clear, actionable error messages through the API rather than unhandled exceptions that leave the system in an unknown state.
+
+---
+
+#### Section 6 - Key Technical Decisions and Why?
+
+**Why Apache MINA SSHD over JSch?** MINA SSHD is the library Ansible and many enterprise Java tools use. It supports modern key types (Ed25519), has async APIs, and is Apache-licensed with active maintenance. JSch is effectively abandoned.
+
+**Why Docker over standalone Debezium Server JAR?** Debezium Server is available as a configurable, ready-to-use application. Using its container image guarantees that the same exact runtime runs identically whether the host is a bare metal server in a data center or a cloud VM. [Debezium](https://debezium.io/) It also makes cleanup trivial - removing a container removes everything.
```
kmos: Please remove mentions of Docker, as it is subject to licensing. My suggestion is to refer instead to a general “container engine.”

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,810 @@
...
...oyment flow: how `PipelineService` fires outbox events, how `ConductorEnvironmentWatcher` picks them up via the embedded Debezium Engine, how `OutboxParentEventConsumer` delegates to `PipelineConsumer`, and how `PipelineConsumer` calls `EnvironmentController.pipelines().deploy()`. This is the exact flow my `HostEnvironmentController` will plug into. <br><br> - Trace the existing `OperatorPipelineController` and `PipelineMapper` code to understand how a `PipelineFlat` gets translated into a `DebeziumServer` CRD - the config prefix resolution (`database.` for sources, `producer.` for Kafka sinks), signal channel setup, and offset/schema-history configuration. My `HostPipelineMapper` will need to produce equivalent `application.properties` from the same input. <br><br> - Set up the full local development environment <br><br> - Schedule a kick-off discussion with mentors to align on two key design questions: (1) how should `PipelineService.environmentController()` route pipelines to the correct `EnvironmentController` (currently hardcoded to `getFirst()`), and (2) should the `HostTarget` credentials integrate with the existing `VaultEntity` system or use a separate secrets mechanism. |
```
kmos, line 717: Can you use another format for this section ? It's difficult to read it

d1vyanshu-kumar: Yes, the table renders properly in the preview/display rich diff mode.
Please check out this link: 
https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/6abdb5c3fc19499540e76996ef020ebebfe3d7b6/gsoc/2026/d1vyanshu-kumar/proposal.md#5-potential-timeline-for-the-gsoc-period

You can also review it in Google Docs. I originally created it there and then converted it to Markdown, which might have caused some issues.
Here’s the Google Doc link:

https://docs.google.com/document/d/1ql_l3J9bLcH8ecCzNXNWUjLP5Ou9ELYR4L6Lfqp02Qs/edit?usp=sharing

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: Thank you for the review. I will fix this issue.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: Good point, thank you! I was overcomplicating it. The platform already has VaultEntity for managing secrets - Sources and Destinations use it for database passwords, so I'll integrate SSH credentials into that same system (HostTarget references a Vault entry with the private key or password). For simpler setups, a keyring integration would be a good, lightweight alternative.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: I already shared in the contribution section at the end of the proposal.
Here is the link:

https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/6abdb5c3fc19499540e76996ef020ebebfe3d7b6/gsoc/2026/d1vyanshu-kumar/proposal.md#my-open-source-contributions

d1vyanshu-kumar: Although the Feedback link isn’t working since it’s private, I’ll make sure to submit the screenshot when I upload the proposal to the GSoC portal.

d1vyanshu-kumar:
Hi @kmos, thank you so much for your feedback. I’ve made some changes. Could you please review them again and let me know if they look good?

I also wanted to mention that the timeline formatting looks a bit off in code view, but if you check it in preview mode, it appears much better. If it still doesn’t look good to you, I’ll be happy to adjust it.

Thanks again for your time and support!

mfvitale (approved):
I @d1vyanshu-kumar it LGTM. Let's wait for @kmos pending points.

gsoc/2026/d1vyanshu-kumar/proposal.md
```diff
@@ -0,0 +1,817 @@
...
+
+    String stdout = readStream(channel.getInvertedOut());
+    String stderr = readStream(channel.getInvertedErr());
+    int exitCode = channel.waitFor(EnumSet.of(ClientChannelEvent.CLOSED), EXEC_TIMEOUT);
+
+    return new CommandResult(stdout, stderr, exitCode);
+}
+```
+
+**Credential Security:** SSH credentials must never be stored in plain text. The platform already has a Vault system (`VaultEntity`) that stores secrets as key-value pairs with a plaintext flag. Each Source, Destination, and Transform can reference multiple Vaults. For host-based deployments, I'll integrate SSH credentials into this existing Vault system - the `HostTarget` entity will reference a Vault containing the SSH private key or password. This approach keeps credential management consistent with how the platform already handles database passwords and API keys - through the same Vault abstraction. The `HostVaultController` (implementing the existing `VaultController` interface) will handle deploying vault contents to the host environment in a secure manner.
```
mfvitale, line 381: @d1vyanshu-kumar  Actually the `Vaults` are not implemented in the platform. So effectively there is not need to support them on the hosts. 

On the other end we need a way to store the ssh keys for the SSH connection. Could you propose an approach for this?

gsoc/2026/d1vyanshu-kumar/proposal.md
kmos: Thanks for the help!

gsoc/2026/d1vyanshu-kumar/proposal.md
kmos: I was interested in Jenkins which I don't see any link.

kmos (approved):
Thanks @d1vyanshu-kumar . I was curious about Jenkins but LGTM the proposal. In my opinion you can go head with a official proposal.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: You're right, I apologise for the confusion. I looked more closely at the `VaultController `and `OperatorVaultController `code and can see that the deploy() and undeploy() methods are empty, the vault system has the data model scaffolded, but the actual deployment logic isn't implemented yet. I shouldn't have assumed it was functional.

For SSH credential storage, here's what I'd propose:

The simplest approach would be to store the SSH private key and password directly in the HostTargetEntity, encrypted at rest. I'd add an encrypted text field for the private key and use a straightforward AES encryption 
utility, the encryption key itself would come from Quarkus configuration (injected via environment variable in production, so it never sits in source code or plain config files). This keeps it self-contained with no dependency on external infrastructure.

Alternatively, as @kmos suggested, a system keyring integration (via Java's KeyStore API) could work well for setups where storing credentials in the database isn't desirable, the key material stays in the OS-level keyring, and the database only holds a reference to the keyring entry.

I'd lean toward the encrypted database column for the initial implementation since it requires no additional infrastructure, and the keyring approach could be added as an alternative backend later. Happy to discuss which direction you'd prefer.

gsoc/2026/d1vyanshu-kumar/proposal.md
d1vyanshu-kumar: @kmos, I had made some changes in Jenkins, but at that time I was caught up with my semester exams and couldn’t complete the PR. The admin eventually closed it.

Here are the links for reference:
• Issue: https://github.com/jenkinsci/jenkins/issues/23888
• PR: https://github.com/jenkinsci/jenkins/pull/25944

d1vyanshu-kumar:
Thanks a lot @mfvitale @kmos for your approvals and review. I’ll now submit the proposal on the official GSoC website.

gsoc/2026/d1vyanshu-kumar/proposal.md
mfvitale: Another alternative could be to let users store keys in kubernetes secret and mount those secret to conductor pod. @d1vyanshu-kumar WDYT?

d1vyanshu-kumar: Hmm...That's a really clean approach and simpler than what I proposed. The user creates a Kubernetes Secret with the SSH private key, mounts it into the Conductor pod as a volume, and the HostTargetEntity just stores the file path where the key is mounted. The actual key material never touches the database.

From the SSH layer side, this maps naturally; the SshSessionManager already loads a key pair from a path reference, so it would just read from the mounted file instead of decrypting from the database. And 
Kubernetes handles encryption at rest and access control natively, so there's no need for custom encryption logic on our end.

I think this is the right approach. I'll update the proposal to reflect it.

Thanks! @mfvitale

PR #34, opened 2026-03-29
Add GSoC 2026 proposal for host-based pipeline deployment

VanKhanhAnny:
Hi,

This PR adds my GSoC 2026 draft proposal for Host-Based Pipeline Deployment for Debezium Platform.

I would appreciate any feedback on the MVP scope, technical direction, and roadmap.

Thanks.


gsoc/2026/VanKhanhAnny/proposal.md
```diff
@@ -0,0 +1,386 @@
...
+
+#### 6. Container-engine-first deployment path
+
+From the official project description, I think a container-based runtime is the right place to start.
+
+For the MVP, the deployment flow would be:
+
+1. Validate the target host and runtime prerequisites.
+2. Render Debezium Server configuration from the pipeline model.
+3. Transfer required artifacts to the target host.
```
kmos: Could you be more specific here? In your approach, it isn’t clear what should be transferred to the target machine.

kmos (commented):
Thank you for the contributions and interest in Debezium. 

My final suggestion for the proposal is to include more technical details that demonstrate both the feasibility of the project and your capabilities.

gsoc/2026/VanKhanhAnny/proposal.md
```diff
@@ -0,0 +1,386 @@
...
+Content-Type: application/json
+```
+
+```json
+{
+  "name": "analytics-vm-1",
+  "hostname": "10.0.0.25",
+  "port": 22,
+  "username": "debezium",
+  "credentialRef": "vault:ssh-host-key",
```
mfvitale: Actually `Vaults` are not yet implemented. Could you propose a different way to manage the keys?

gsoc/2026/VanKhanhAnny/proposal.md
```diff
@@ -0,0 +1,386 @@
...
+My expectation is that the MVP needs a host-target registration path in Conductor, likely exposed through backend APIs first. A user should be able to register a host target with:
+
+- human-readable name
+- hostname or IP address
+- connection port
+- username or connection identity
+- credential or secret reference
+- optional labels or environment metadata
+
+After registration, Conductor should be able to run preflight checks before a pipeline is associated with that host. That association then determines whether deployment uses the operator-backed environment or the host-based environment.
```
mfvitale, line 165: What about configuring the host to be able to run DS (i.e. install a container runtime)?

mfvitale:
Hi @VanKhanhAnny thanks for your proposal. It LGTM at high level, I left some comments.

gsoc/2026/VanKhanhAnny/proposal.md
VanKhanhAnny: I updated this part to be more specific. For the container-engine MVP, I now describe transferring a small runtime bundle to the target host such as a rendered `application.properties`, launch metadata (env file or a small launch script), and the working-directory layout for generated files, logs, and lifecycle metadata. The Debezium Server image itself would be pulled on the host by the container engine rather than copied from Conductor

VanKhanhAnny:
> Thank you for the contributions and interest in Debezium.
> 
> My final suggestion for the proposal is to include more technical details that demonstrate both the feasibility of the project and your capabilities.

Thanks for the feedback. I updated the proposal to add more technical detail around the current Conductor deployment flow, the host-target model, the preflight/bootstrap path, and the container-engine deployment flow so the MVP is more concrete and easier to evaluate

gsoc/2026/VanKhanhAnny/proposal.md
VanKhanhAnny: Thanks, I updated the proposal to remove the vault-based assumption. Since vault support is not implemented yet, I now propose that Conductor receives SSH private keys through its own runtime environment, for example as mounted files or deployment-time secret volumes, and that the host model stores only a reference to that mounted key

gsoc/2026/VanKhanhAnny/proposal.md
VanKhanhAnny: Good point. I updated the proposal to make that explicit. The revised version now includes host preflight checks for SSH access, runtime presence, working-directory setup, and image-pull capability. For one supported Linux family, I also added a narrow bootstrap path where Conductor can install the container engine, ensure the service is running, and verify that the target user can start a simple test container

gsoc/2026/VanKhanhAnny/proposal.md
kmos: Thanks!

kmos (approved):
Approved, thank you for your contributions and interest in Debezium!

mfvitale (approved):
@VanKhanhAnny Thanks! LGTM! So you can proceed with the final submission!

Thanks again for your effort.

PR #35, opened 2026-03-31
GSoC 2026 proposal: Host-Based Pipeline Deployment

JerryGao0805:
looking forward to comments, Thank you!

### DIFF
--- a/gsoc/2026/JerryGao0805/proposal.md
+++ b/gsoc/2026/JerryGao0805/proposal.md
@@ -1,0 +1,453 @@
+# Debezium: Host-Based Pipeline Deployment for the Debezium Platform
+
+## About Me
+
+1. **Name:** Yang Gao (GitHub: [JerryGao0805](https://github.com/JerryGao0805))
+2. **zulip message:** [Zulip link](https://debezium.zulipchat.com/#narrow/channel/573881-community-gsoc/topic/newcomers/near/582504580)
+3. **University:** Stevens Institute of Technology — M.S. in Computer Science (Dec 2026 graduation)
+4. **Contact:** yanggao0805@gmail.com
+5. **Timezone:** EST (GMT-4)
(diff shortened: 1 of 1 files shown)


gsoc/2026/JerryGao0805/proposal.md
```diff
@@ -0,0 +1,453 @@
...
+| `GET` | `/api/pipelines/{id}/targets` | List deployment targets |
+| `POST` | `/api/pipelines/{id}/targets` | Add a deployment target (type + host ref) |
+| `DELETE` | `/api/pipelines/{id}/targets/{targetId}` | Remove a deployment target |
+
+**Backward compatibility:** Pipelines without any deployment target default to the existing Operator behavior — no migration needed for existing data.
+
+#### SSH Credential Management
+
+**Storage — Java Keystore (JKS / PKCS12):**
+- A keystore file lives on the Conductor host, protected by a master password from an environment variable (`CONDUCTOR_KEYSTORE_PASSWORD`)
```
mfvitale: Could you analize the pro/cons of this solution compared with the approach to use k8s secrets mounted as a volume on the conductor?

mfvitale (approved):
Hi @JerryGao0805 thanks for your proposal and you effort!

I left just one minor comment, once addressed you could proceed with the official submission!

gsoc/2026/JerryGao0805/proposal.md
JerryGao0805: Thanks for your feedback!  
I think Java Keystore works anywhere the JVM runs, doesn't depend on K8s, and gives you encryption at rest out of the box. I initially went with it since I've used it in previous projects and it felt like a safe and good choice.    
K8s Secrets is more simpler, there is no custom credential management code, easily fetch from a mounted path. You also get K8s RBAC, hot reloading, and friendly when integrating with other external cloud key management services.
Since the runtime environment is already K8s, using K8s Secrets is probably a more natural way. I think I can probably see K8s Secrets as the default, with a pluggable CredentialProvider interface so Keystore can serve as a non-K8s setups.

PR #37, opened 2026-04-10
DDD-36 Oracle Multithreaded Stream Processing

### LINKED ISSUE #36: Oracle multi-threaded model
Add a new model for the Oracle connector that uses multiple threads to improve LogMiner efficiency. This new model would use a shared buffer mechanism between multiple threads to expedite reads, minimize memory consumption, and improve throughput to Kafka.

Naros:
Fixes #36

### DIFF
--- a/DDD-36.md
+++ b/DDD-36.md
@@ -1,0 +1,266 @@
+# DDD-36: Oracle Multi-threaded Streaming
+
+## Overview
+
+This document describes a proposed redesign of the Debezium Oracle connector's LogMiner processing pipeline.
+The current implementation is single-threaded, combining database reading, event processing, and event dispatch into a sequential flow, which we've traditionally used for all our connectors.
+But this approach for LogMiner leads to heap pressure from large in-flight transactions and limits throughput by coupling IO bound and CPU bound work together.
+
+The proposed architecture splits the pipeline into three logical threads, connected by two `ChronicleQueue` instances, both using disk-backed memory-mapped files for inter-thread communication.
+This eliminates heap exhaustion for large transactions while improving throughput, observability, and scalability.
+
+## Current Architecture
+
+The existing single-threaded flow is:
+
+1. Execute LogMiner SQL, iterate `V$LOGMNR_CONTENTS` results over a JDBC `ResultSet`.
+2. Map each row to a `LogMinerEventRow` POJO, parse the associated `SQL_REDO`, and cache the POJO's metadata and parsed `SQL_REDO` as a `LogMinerEvent` object in the transaction buffer.
+3. On `COMMIT`, pop the transaction off the buffer, apply merge/discard logic, and emit.
+4. On `ROLLBACK`, pop the transaction off the buffer, discarding the entire transaction and event list.
+
+The primary limitations of this approach are:
+
+* **Heap Exhaustion**: Large or long-running transactions accumulate unbounded heap usage as events are buffered until a `COMMIT` or `ROLLBACK` is observed. 
+While the connector does have off-heap buffer options using Infinispan and Ehcache, the disk-based persistence offered by these solutions is slow, and contributes to throughput issues.
+* **Coupled I/O**: The database reader stalls while processing and emitting events, reducing the LogMiner query throughput.
+* **No audit trail**: All LogMiner output is discarded after processing, with no ability to replay and/or inspect what we originally mined.
+This becomes even more important as Oracle often only retains logs for a limited period of time (sometimes hours).
+
+## Proposed Architecture
+
+### High-Level Flow
+
+![High-Level Design](DDD-36/oracle_multiple_thread_high_level.svg)
+
+Each thread is fully decoupled from the others.
+Threads 1 and 3 are purely I/O-bound.
+Thread 2 owns all the processing logic.
+The two `ChronicleQueue` queue instances serve as durable, disk-backed inter-thread transports, replacing heap or other off-heap solutions.
+
+### Thread 1 - Reader
+
+Thread 1 is a deliberately simple, fast database reader.
+Its only job is to execute the LogMiner SQL, iterate the `ResultSet`, mapping each row to a structured entry written to Queue 1.
+It contains no business logic.
+
+#### Behavior
+
+* Executes LogMiner SQL against `V$LOGMNR_CONTENTS`
+* Maps each JDBC row to an entry in Queue 1 (see below)
+* Tracks Queue 1's last written high watermark (TBD) to avoid duplicate writes on restarts or re-reads
+* Returns to querying the next LogMiner batch as fast as possible
+
+#### High-watermark tracking
+
+Because LogMiner re-reads from the eldest in-progress transaction's SCN, Thread 1 will often re-observe portions of transactions it has already written to Queue 1.
+The `last_written_watermark` is persisted in Debezium's offset and used to skip rows already present in Queue 1.
+
+```java
+if (logMinerEventRow.getPosition() > lastWrittenWatermark) {
+  writeRowToQueue(logMinerEventRow);
+  updateLastWrittenWatermark(logMinerEventRow);
+}
+```
+
+### Queue 1 - LogMiner Raw Data
+
+Queue 1 is an immutable, append-only audit log of everything Oracle LogMiner has produced.
+It serves two primarily roles:
+
+1. **Inter-thread transport**: Thread 2 will read from it to process transactions.
+2. **Audit Log**: This can be retained for debugging purposes or the ability to replay events.
+
+#### Entry Schema
+
+Each entry in the queue contains a set of structured uncompressed metadata key/value pairs, followed by the `SQL_REDO` payload.
+
+The following are the uncompressed metadata fields.
+These fields are not compressed as they're read by Thread 2 to maximize Chronicle Queue pre-processing pass speeds.
+
+| Field            | Type     | Notes                                                          |
+|------------------|----------|----------------------------------------------------------------|
+| `XID`            | `String` | Transaction identifier                                         |
+| `SCN`            | `String` | System change number                                           |
+| `OPERATION_CODE` | `int`    | Oracle's numeric operation code                                |
+| `THREAD#`        | `int`    | Assigned redo thread number                                    |
+| `TABLE_NAME`     | `String` | The fully qualified `TableId` value, serialized as a `String`  |
+| `ROW_ID`         | `String` | The Oracle unique table row identifier assigned to the change  |
+| `CSF`            | `int`    | Continuation flag (1=next row merge with this row, 0=last row) |
+| `ROLLBACK`       | `int`    | Savepoint/Partial rollback flag                                |
+| `RS_ID`          | `String` | Rollback segement identifier                                   |
+| `SSN`            | `int`    | SQL sequence number                                            |
+
+The following are the compressed fields.
+
+| Field      | Type     | Notes                                        |
+|------------|----------|----------------------------------------------|
+| `SQL_REDO` | `String` | The raw Oracle reconstructed SQL, compressed |
+
+#### Compression
+
+The `SQL_REDO` column is the dominant contributor to Queue 1 disk usage.
+Oracle's reconstructed SQL is highly repetitive (table names, column names, and so on).
+When this is repeated across multiple rows, this compresses quite well.
+Compression is applied at the field level, leaving all metadata fields uncompressed.
+
+Queue compression will be made configurable:
+
+```java
+public enum SqlRedoCompression {
+    NONE, // fastest write/read, highest disk usage
+    LZ4,  // default - low CPU overhead, good ratio
+    ZSTD, // best ratio, higher CPU and recommended for large transactions
+}
+```
+
+The `LZ4` compression is recommended and the default, while `ZSTD` is preferred in Oracle environments where long-running transactions risk disk exhaustion.
+
+#### Retention / Cleanup
+
+The Queue 1 entries are retained until both of the following conditions are met:
+
+1. The `scn >= queueEntry.SCN` (Thread 1 has advanced past this entry)
+2. The `commit_scn >= queueEntry.SCN` (Thread 3 has emitted the transaction containing the entry)
+
+Cleanup is driven by `StoreFileListener`, which fires when a queue file reaches maximum capacity and rolls.
+When the queue rolls to a new active file, non-active files become eligible for deletion.
+
+A map of "roll cycle to maximum SCN" is maintained by Thread 1, and used to evaluate the deletion condition:
+
+```java
+private boolean isSafeToDelete(int cycle) {
+    final long maxScnInCycle = cycleScnIndex.getMaxScn(cycle);
+    return scnWatermark >= maxScnInCycle && commitScnWatermark >= maxScnInCycle;
+}
+```
+
+### Thread 2 - Processor
+
+Thread 2 owns all transaction processing logic.
+It is responsible for reading from Queue 1, reconstructs transactions by `XID` (transaction id), applies the multiple event consolidation into a single logical event as well as partial rollback discard rules, and writes the final reconstructed set of events for each transaction into Queue 2.
+
+#### Transaction Index Lookup
+
+Thread 2 is responsible for maintaining a lightweight heap map of in-flight transactions:
+```java
+Map<String, Long> transactionIdToBeginIndex; // XID to Queue 1 index of BEGIN events
+```
+
+This map holds only a `long` index per transaction (8 bytes), regardless of how many events the transaction contains.
+The event payloads themselves remain on disk in Queue 1.
+This eliminates the heap exhaustion risk that is present in the current implementation.
+
+#### Processing Flow
+
+When Thread 2 observes a `COMMIT` for a given `XID`, it performs two unique passes over the transaction's entries in Queue 1.
+
+**Pass 1: Build thin index**
+
+Thread 2 seeks to the stored `BEGIN` index for the `XID`, reading the queue forward until the `COMMIT`.
+During this pass, only uncompressed metadata key/value pairs are used.
+This pass builds:
+
+1. `List<ThinIndex>`: a lightweight index of all events in the transaction
+2. `Set<Long> skipIndexes`: Set of indices from Queue 1 for the transaction to discard (e.g. partial rollbacks)
+3. `Map<LobKey, LobGroup>`: Grouping of LOB fragment sequences for merges.
+
+**Pass 2: Assemble**
+
+Thread 2 iterates the `ThinIndex` list and for each retained event, seeks back to its `queueIndex` in Queue 1, decompresses the `SQL_REDO`, and writes the final event to Queue 2.
+Any event in `skipIndexes` are never decompressed.
+
+For LOB merge events, Pass 2 seeks to each fragment's `queueIndex`, decompresses and merges the fragment values.
+This simulates the logic that currently exists inside the `TransactionCommitConsumer` implementation.
+Once all groupings are merged, the consolidated event is written to Queue 2.
+
+**Discard Logic - Partial Rollbacks**
+
+LogMiner emits events with `ROLLBACK=1` flag.
+This most often is when a savepoint is rolled back within a transaction, but can represent other scenarios like constraint violations, where Oracle records the original insert, and then reverts it due to the constraint.
+These events cancel earlier events that often have the same `ROW_ID`.
+Both the original and the rollback event are added to the `skipIndexes`, and are excluded from Queue 2, as well as the decompression of their respective `SQL_REDO` column.
+
+**Rollback Handling**
+
+When Thread 2 observes a `ROLLBACK` for a transaction, the entire transaction is discarded.
+In this case, the `xidToBeginIndex` entry for the transaction's `XID` is removed and nothing is written to Queue 2.
+
+### Queue 2 - Processed Events
+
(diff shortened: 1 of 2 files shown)


DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+In this case, the `xidToBeginIndex` entry for the transaction's `XID` is removed and nothing is written to Queue 2.
+
+### Queue 2 - Processed Events
+
+Queue 2 contains the final, clean set of events to be emitted to the target system.
+It is written exclusively by Thread 2 and read exclusively by Thread 3.
+
+Each transaction in Queue 2 begins with a `BEGIN` event, followed by its events and terminated by a `COMMIT`.
+The `COMMIT` marker carries the `commit_scn` along with an event count.
+Thread 3 waits for this `COMMIT` marker before emitting the transaction.
```
Naros, line 249: Actually, as long as Thread 2 writes all necessary markers into Queue 2, Thread 3 could actually start emitting the transaction as soon as Thread 2 begins to write the entries, as the entries will be processed and written in the transaction commit, user-executed event order.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+
+#### Entry Schema
+
+Queue 2 entries are fully processed, structured events with parsed column values and no raw SQL, unless the user has configured to include the raw SQL.
+If the SQL is to be included, it will be compressed.
+The exact schema to be used here generally will align to contain all needed fields to populate the `source` information block and the `before` and `after` portions of a Debezium event.
+
+### Thread 3 - Emitter
```
Naros, line 257: It might be useful to also highlight that because Thread 3 is responsible for the emission of the events to the `EventDispatcher`, this is also where the `ReselectPostProcessor` logic fires, and can create even more back-pressure to the streaming pipeline. The separate thread avoids this back-pressure on either of the two former threads.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
+# DDD-36: Oracle Multi-threaded Streaming
+
+## Overview
+
+This document describes a proposed redesign of the Debezium Oracle connector's LogMiner processing pipeline.
+The current implementation is single-threaded, combining database reading, event processing, and event dispatch into a sequential flow, which we've traditionally used for all our connectors.
+But this approach for LogMiner leads to heap pressure from large in-flight transactions and limits throughput by coupling IO bound and CPU bound work together.
```
kmos: Can this redesign be enabled via an option? I was thinking that, in some scenarios, some users might prefer the single-threaded solution.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+It contains no business logic.
+
+#### Behavior
+
+* Executes LogMiner SQL against `V$LOGMNR_CONTENTS`
+* Maps each JDBC row to an entry in Queue 1 (see below)
+* Tracks Queue 1's last written high watermark (TBD) to avoid duplicate writes on restarts or re-reads
+* Returns to querying the next LogMiner batch as fast as possible
+
+#### High-watermark tracking
```
kmos, line 95: Can you add a section with references related to the cited technologies?

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+* **One entry per row**: The `writeDocument` API must be called once per `ResultSet` row, not once per batch. 
+Wrapping multiple `ResultSet` rows in a single `writeDocument` will exhaust the mapped chunk boundary, and throw a `DecoratedBufferOverflowException`.
+* **Roll Cycle**: hourly rolling with aggressive cleanup keeps disk usage bounded, subject to long-running transactions.
+* **Block Size**: A 64MB block size by default. This should be tunable based on observed average entry size.
+
+### Concerns
+
+`ChronicleQueue` does come with its own sets of concerns, most notably the requirement to set the following JVM arguments due to how the library uses various Sun-specific APIs:
```
kmos: About this point (even if it’s an implementation detail), what do you think about reducing coupling by wrapping `ChronicleQueue` behind an indirection layer? That way, if we want to change or switch to a different technology, it would be easy to implement.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+Thread 2 is responsible for maintaining a lightweight heap map of in-flight transactions:
+```java
+Map<String, Long> transactionIdToBeginIndex; // XID to Queue 1 index of BEGIN events
+```
+
+This map holds only a `long` index per transaction (8 bytes), regardless of how many events the transaction contains.
+The event payloads themselves remain on disk in Queue 1.
+This eliminates the heap exhaustion risk that is present in the current implementation.
+
+#### Processing Flow
```
kmos, line 207: If I understand correctly, in the `Processor` are also applied the transformers right? I’m not sure in which phase this happens.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+#### Behavior
+
+* Executes LogMiner SQL against `V$LOGMNR_CONTENTS`
+* Maps each JDBC row to an entry in Queue 1 (see below)
+* Tracks Queue 1's last written high watermark (TBD) to avoid duplicate writes on restarts or re-reads
+* Returns to querying the next LogMiner batch as fast as possible
+
+#### High-watermark tracking
+
+Because LogMiner re-reads from the eldest in-progress transaction's SCN, Thread 1 will often re-observe portions of transactions it has already written to Queue 1.
```
vjuranek: Can it be reduced by using `AS OF SCN` to select only records we haven't processed yet?

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Thread 3 - Emitter
+
+Thread 3 is purely an IO-bound producer thread.
+It reads all processed events from Queue 2 and emits them as `SourceRecord` instances.
+All events are emitted from Queue 2 in transaction commit chronological order.
+
+The `commit_scn` represents the highest transaction's commit point for a given Oracle redo thread.
+Thread 3 is responsible for maintaining this offset state.
+
```
vjuranek, line 265: It may be useful to track the progress of emitting the records with `lastEmitWatermark` (similar to `lastWrittenWatermark`) if where is any failure of immediate stop, so we can resume from the last emitted record and don't duplicate records already sent from the beginning of the transaction.

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+The library authors are already working on newer releases not to require these without any loss of performance.
+
+## Scalability
+
+### Multiple Kafka Connect Tasks?
+
+By utilizing multiple threads to handle independent stages during streaming with inter-thread concurrent transport, a given stage could be promoted to a Kafka Connect task instead.
+By using multiple tasks, this would allow the connector to take advantage of a Kafka Connect cluster's full processing power, similar to how Google Spanner works.
```
vjuranek, line 421: This would require shared file system across multiple machines, which is not supported by Chronicle, see Important note in [this section](https://github.com/OpenHFT/Chronicle-Queue/#usage).

DDD-36.md
Naros: I've definitely thought about this, as well as whether or not making this pluggable would be the right solution, which allows us to integrate a portion of [Nathan's PR](https://github.com/debezium/debezium/pull/6820) work to introduce beyond Infinispan, Ehcache, and CQ, but RocksDB too.

From a code reorganization and minimizing technical debt, I do think having a single-threaded mode makes sense, and if we truly can make the "queue" that sits between these stages pluggable, then having a single-threaded mode gives us the ability to have a single architecture while preserving a lot of the existing connector behavior, while at the same time offering a "performance mode".

DDD-36.md
Naros: That's definitely the idea, and I agree. Nathan has already done this using our current transaction buffer API, so we can certainly build on that idea. 

The only thing I would change about his design is that he used a queue/transaction; for performance, a single queue for all works better from system ulimit and file-handle concerns, given that CQ relies heavily on memory-mapped files. But this is just an implementation detail.

DDD-36.md
Naros: Just to be clear, this does not invoke or relate to Kafka's `Transformation` API. 

Today the `AbstractLogMinerStreamingChangeEventSource` performs several in-flight processing steps for every row read from LogMiner (consolidating rows with CSF=1, parsing the SQL, examining metadata), and then performs a post-pass of processing steps when we see the `COMMIT` for the transaction (merging LOB fragments, consolidating multi-stage SQL operations for a logical change event) from the `TransactionCommitConsumer` class.

The processor is effectively the successor to this logic.  It shifts to a bulk operation where we do both the in-flight processing and the post-pass processing as post-commit steps. This brings several performance benefits, namely the ability to avoid parsing or performing any of the in-flight processing steps on every row for transactions that are discarded, e.g. rollbacks, partial rollback rows, and others. In short, we only do the expensive processing on the rows that matter, for the cost of some disk space.

The Kafka `Transformation` stage happens after Thread 3 has handed the event off to the `EventDispatcher` and the runtime (Debezium Server, Embedded Engine, Kafka Connect) have pulled a subset of queued events off the internal `ChangeEventQueue`.

DDD-36.md
Naros: We can't use the Flashback Area query semantics here, but I've considered having Thread 1 maintain a list of transactions we've already seen commits/rollbacks for and excluding those from the query. 

My concern is that, with Oracle logging everything, there can easily be thousands of transactions if you have a long-running transaction. While this might help avoid bandwidth overhead sending the data to Debezium only to discard, I'm not yet sure whether applying that at the query level helps or hurts. It's the same argument we've seen wrt the `log.mining.query.filter.mode`. In some environments, it works well, and others not so much to the point it makes it worse :sob: 

These types of situations I want to avoid if at all possible.

DDD-36.md
Naros: Thanks for pointing that out. I hadn't seen that yet, but that's not too surprising that being an Enterprise Feature. I do think if we circle back to the point above that Giovanni and I were discussing with single vs multi-threaded toggles, if we make the API pluggable, perhaps network-based solutions like an Infinispan Server deployment would fit well in this model.

The big question I have is what is the net gain, e.g., gains offset by the complexity?

DDD-36.md
Naros: Just so I understand, are you referring to a section that links to things like CQ, ISPN, Ehcache, RocksDB, etc?

kmos: I mean for example [Oracle High-watermark](https://docs.oracle.com/en/database/oracle/oracle-database/19/cncpt/Chunk806120860.html#GUID-F3F5ACB2-94C8-483F-B427-1981836D0609)

DDD-36.md
kmos: Thanks a lot for the clarification!

DDD-36.md
```diff
@@ -0,0 +1,266 @@
+# DDD-36: Oracle Multi-threaded Streaming
+
+## Overview
+
+This document describes a proposed redesign of the Debezium Oracle connector's LogMiner processing pipeline.
+The current implementation is single-threaded, combining database reading, event processing, and event dispatch into a sequential flow, which we've traditionally used for all our connectors.
+But this approach for LogMiner leads to heap pressure from large in-flight transactions and limits throughput by coupling IO bound and CPU bound work together.
+
+The proposed architecture splits the pipeline into three logical threads, connected by two `ChronicleQueue` instances, both using disk-backed memory-mapped files for inter-thread communication.
```
rk3rn3r, line 9: Maybe those queues could be made configurable? Do you think a Kafka topic, Redis queue, etc could be a good fit? Would it then make sense to decouple the reader from the processing layer entirely and have different executables for those different use cases?

DDD-36.md
```diff
@@ -0,0 +1,266 @@
...
+This eliminates heap exhaustion for large transactions while improving throughput, observability, and scalability.
```
rk3rn3r: How is observability improved? When this is spread over different threads or even processes I would say things get harder. Do events receive a unique identifier for tracing?

DDD-36.md
Naros: I've actually thought about that, but the main limitation I see is the lack of random access and ordering.

With a Kafka topic, the only way to achieve what ChronicleQueue supports is with a single-partition topic. If you are pushing millions of changes into that topic with only a single partition, I quickly think Kafka will become a bottleneck. Aside from the throughput concerns, I think it fits well otherwise.

For Redis, a Redis Stream might work, as we could also read from an offset position sequentially, but the bigger concern here is data volume and memory pressure. 

For raw performance, CQ will outperform these two approaches N-fold.

DDD-36.md
Naros: Most often, when issues are detected, the archive/redo logs are already purged. 

With the CQ queue, the user could have a longer alternative retention policy for the change events that we could go back to and use if we need to identify why a specific issue happened or if it was related to LogMiner, etc.

DDD-36.md
rk3rn3r: Could you elaborate on the point of `random access and ordering` ?  I guess you mean you require total ordering of events? Ordering per PK is not enough?

DDD-36.md
rk3rn3r: So the longer durability that gives you better trace-ability?

rk3rn3r:
A general note on CQ: CQ performance sounds great, but is there a proof that CQ will work with huge messsages like the ones from a db replication log? The use case they describe sounds like typical IPC use case. I could imagine that it works great as IPC queue but the concept could be vulnerable when used with typical Debezium workloads that tend to be huge in comparison.

DDD-36.md
Naros: > Ordering per PK is not enough?

No, it is not. For transaction processing, Oracle requires that we maintain the precise user-execution order of all events within a single transaction. This is because the transaction logs write committed, uncommitted, rolled-back, and partially rolled-back events to the same event stream, and without preserving order, we won't be able to handle all use cases.

For example, when a partial rolled-back event occurs, we need to be able to step through the transaction buffer in reverse from the current event, identify the event with the matching ROW_ID (not PK), and then discard both the partial rolled-back event and the match. 

The only way to do that in a Kafka topic is a keyless, single-partition topic.

In terms of random access, the goal is to use Queue 1 to store a replica of the Oracle transaction log, but only containing the events the connector is interested in. This queue acts as a replacement for using heap, infinispan, or ehcache storages. When a COMMIT is observed, we notify thread 2 to process that transaction, and it will seek to the start of that transaction and read from that position to the COMMIT.

So we need random seek to an offset and the ability to read from that position forward.

Naros:
@rk3rn3r, the IPC use case is for the lowest-latency use case. But as I outlined in the Q1 layout, a single event will range from just a few hundred bytes to 1k-4k, depending on how efficient LZ4 compression is for the SQL column data. Those event payloads are well within CQ's wheelhouse to handle without any memory or I/O issues, even if we were to be writing millions of events per second.

DDD-36.md
mfvitale: I was thinking the same, since if I got it correctly in the `Queue 2` we only have committed transactions events so there is not possibility for a record to be reverted.

mfvitale:
I @Naros This is a great enhancement that, I can imagine, could be also extended to other connectors to right? 

There is a thing that is not completely clear to me. We have the T1 that do the redo log stuff and put raw data into the Q1 that is, more or less, a shadow copy of the transaction log. This is the first point of decoupling. 
So how the offset is manages here after a connector failure? Currently, all is coupled in an unique pipeline and the only de-coupling between the record writes and their offsets it on the KC level. So this is why we generally have the `at-least-once` delivery. 

It is not cleat to me how the T1 restore from a connector failure.

Naros:
> I can imagine, could be also extended to other connectors to right?

@mfvitale, it absolutely could. We already have other connectors that handle buffering on their own, e.g., Informix, which uses a heap-based approach. Eventually, OOM may be an issue for larger transactions. When ever we finish the work on supporting PostgreSQL PgOutput V2, we'll need to be able to buffer in-flight transactions, too.

> So how the offset is manages here after a connector failure? Currently, all is coupled in an unique pipeline and the only de-coupling between the record writes and their offsets it on the KC level. So this is why we generally have the at-least-once delivery.

So this is probably something worth adding to the DDD, so thanks for the reminder.

Oracle's offsets are already somewhat unique in that they maintain multiple layers of critical detail for LogMiner, so we know how to resume after a prior shutdown or crash (`scn`, `commit_scn`, `txId`, and `txSeq`).

So today, when the connector restarts from a crash:

1. It resumes from `scn`, this is the position where the eldest in-flight transaction that was previously buffered starts. The connector begins rebuffering the eldest in-flight transaction, effectively buffering all the transactions that were in the buffer previously.
2. During a resume or remine pass, events are compared against `commit_scn`, and if they have positions that come before the `commit_scn`, those events are discarded. This avoids us rebuffering and reprocessing transactions we've already sent to Kafka.
3. During the commit of a transaction, all events for the transaction are emitted. Each event is assigned its current `txId` (transaction id) and `txSeq` (transaction position). _Transaction position for LogMiner is synthetic because uncommitted mode provides no stable sequence otherwise_. This allows us on restart that if we're reading the same transaction as `txId`, it won't yet be in the `commit_scn` until its fully sent, and we can partially discard the transaction events, so we partially only send the portion that wasn't previously delivered before the crash.
4. Once all events are sent, the `commit_scn` is updated so we know what we've already fully processed.

In the multi-threaded model, it would work very similarly.

1. T1 resumes reading from `scn` on restart.
2. Queue1 position on restart is recorded, and anything in Queue1 from before the crash is discarded. So, on the resume, T1 effectively primes Queue1, just as it primes the transaction buffer today.
3. T2 is a pure processing thread, with the potential for fan-out concurrent processing of transactions. Because ChronicleQueue allows for multi-reader, single-writer operations, we could fan out the preprocessing passes per transaction across a worker thread pool, and once the preprocessing pass data is gathered, jobs are reassembled in transaction commit order for writes to Q2.
4. T3 reads Q2 and emits the events, and is the single thread that writes offset data. This effectively moves the `handleCommitEvent` handler we have today into its own thread, so that neither T1 nor T2 is bound by commit back-pressure, especially for very large transactions with 10s/100s of millions of changes.

Now, in the long term, I would prefer that Q1 not only be a full shadow copy of the transaction logs with a connector-controlled retention policy, but that the connector resumes from the last event in the shadow copy rather than from `scn`, which could be some minutes or even hours earlier. For some connectors, that's certainly doable, but for LogMiner in particular, there are some technical constraints with transaction boundaries that unfortunately Oracle enforces :(.

mfvitale:
> Now, in the long term, I would prefer that Q1 not only be a full shadow copy of the transaction logs with a connector-controlled retention policy, but that the connector resumes from the last event in the shadow copy rather than from `scn`, which could be some minutes or even hours earlier.

I was thinking that the current implementation was already doing in this way and so my question about the offsets. 

So with the current implementation all works as expected since the offsets (scn) guides and the Q1 is then evicted based on it on restore. 

On long term the Q1 will be let's say immutable and for the connector will be the `source` of the events and then the connector offset will be related to the Q1 processing. 

If we then move in this direction I can imagine that we could have a sort of lightweight process that to the transaction log processing and just store events on the Q1, then our connectors will just read from that queue.

Naros:
> On long term the Q1 will be let's say immutable and for the connector will be the `source` of the events and then the connector offset will be related to the Q1 processing.

In this case, I don't think the offset logic changes. You would still want `scn` to point to the oldest in-flight transaction if the user deletes Q1. For the long-term resume without re-processing, Q1 could be tailed by a reader before we begin reading from the database, to identify the last X records and determine where the database read should resume from as an intermediate offset resume position.

> If we then move in this direction I can imagine that we could have a sort of lightweight process that to the transaction log processing and just store events on the Q1, then our connectors will just read from that queue.

That's correct, at least those that need buffering capabilities. The biggest issue I see in Oracle, and why even T3 exists, is that we frequently see where the `ChangeEventQueue` creates far more back-pressure than connectors that stream directly. This is a side effect of buffering large transactions and releasing them all at once, creating a flood effect on the `ChangeEventQueue`.

For connectors that can safely read from the transaction log and emit directly, I would say these are probably better left as the status quo and let users adjust back pressure on `QueueRemainingCapacity` as they do today.

And in fact, thinking back to https://github.com/debezium/debezium/pull/6468, I now wonder whether T3 and Q2 could be eliminated and a specialized `ChangeEventQueue` could be a replacement? :thinking: wdyt?

mfvitale:
> And in fact, thinking back to https://github.com/debezium/debezium/pulls/8968, I now wonder whether T3 and Q2 could be eliminated and a specialized `ChangeEventQueue` could be a replacement? 🤔 wdyt?

The link seems not working..

Naros:
@mfvitale fixed the link :heavy_check_mark:

DDD-36.md
Naros: I believe that `OffsetScn` and `CommittedScn` in the JMX metrics provide this already, right?

mfvitale:
> And in fact, thinking back to https://github.com/debezium/debezium/pull/6468, I now wonder whether T3 and Q2 could be eliminated and a specialized ChangeEventQueue could be a replacement? 🤔 wdyt?

Maybe yes.

mfvitale:
@Naros Effectively the Q1 maps with current transaction buffer?

DDD-36.md
vjuranek: yes, you are right

vjuranek (approved):
LGTM, great stuff @Naros !

mfvitale:
> I now wonder whether T3 and Q2 could be eliminated and a specialized ChangeEventQueue could be a replacement? 🤔 wdyt?

I think so.

PR #39, opened 2026-04-28
DDD-38 Pipeline Monitoring for Debezium Platform

### LINKED ISSUE #38: Debezium Platform monitoring
This will describe how to integrate monitoring inside Debezium Platform

mfvitale:
fixes: #38

### DIFF
--- a/DDD-38.md
+++ b/DDD-38.md
@@ -29,17 +29,18 @@ This represents a significant operational burden and increases the barrier to ad
 3. **Real-time monitoring** - Display current and historical metrics through efficient polling mechanisms
 4. **Simple deployment** - Monitoring infrastructure deployed alongside the platform with minimal configuration
 
-## Proposed Solution: JMX + Prometheus + Custom REST API
+## Proposed Solution: OpenTelemetry + Prometheus + Custom REST API
 
 ### Overview
 
-The monitoring system consists of three key components working together:
+The monitoring system consists of four key components working together:
 
-1. **JMX Exporter** on Debezium Server instances to expose metrics in Prometheus format
-2. **Prometheus** (installed or external) to scrape and store time-series data
-3. **Custom REST API** in Conductor to query metrics and serve them to Stage UI with caching
+1. **OpenTelemetry Java Agent** on Debezium Server instances to collect metrics from JMX MBeans
+2. **OpenTelemetry Collector** to receive, process, and export metrics
+3. **Prometheus** (installed or external) to scrape and store time-series data from the Otel Collector
+4. **Custom REST API** in Conductor to query metrics and serve them to Stage UI with caching
 
-This approach provides end-to-end observability from metric collection to visualization while maintaining control over user experience.
+This approach provides end-to-end observability from metric collection to visualization while establishing OpenTelemetry as the foundation for future native instrumentation. The architecture enables a gradual migration path from agent-based collection to OpenTelemetry SDK native instrumentation.
 
 ### Architecture
 
@@ -58,12 +59,14 @@ graph TB
         Service --> Client
     end
     
-    Prom["Prometheus<br/>- Scrapes metrics<br/>- Stores time series"]
-    DBZ["Debezium Server + JMX Exporter<br/>- Exposes JMX metrics<br/>- Port 9090"]
+    Prom["Prometheus<br/>- Scrapes metrics endpoint<br/>- Stores time series"]
+    OtelCol["OpenTelemetry Collector<br/>- Receivers: OTLP<br/>- Exporters: Prometheus"]
+    DBZ["Debezium Server + Otel Agent<br/>- Collects JMX MBeans<br/>- Exports to OTLP"]
     
     UI1 -->|"HTTP GET /api/v1/monitoring/...<br/>Cache-Control, ETag"| Conductor
     Client -->|"PromQL queries"| Prom
-    Prom -->|"Scrape (HTTP)"| DBZ
+    Prom -->|"Scrape /metrics (HTTP)"| OtelCol
+    OtelCol -->|"OTLP (gRPC/HTTP)"| DBZ
     
     style UI fill:#e1f5ff,stroke:#0288d1,stroke-width:2px,color:#000000
     style Conductor fill:#fff4e1,stroke:#f57c00,stroke-width:2px,color:#000000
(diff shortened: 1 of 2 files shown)


DDD-38.md
```diff
@@ -0,0 +1,598 @@
...
+4. **Simple deployment** - Monitoring infrastructure deployed alongside the platform with minimal configuration
+
+## Proposed Solution: JMX + Prometheus + Custom REST API
+
+### Overview
+
+The monitoring system consists of three key components working together:
+
+1. **JMX Exporter** on Debezium Server instances to expose metrics in Prometheus format
+2. **Prometheus** (installed or external) to scrape and store time-series data
```
kmos: Why not opentelemetry? Prometheus offers metrics but with opentelemetry you have also trace and logs

mfvitale: I did't had a look deeply in the `OpenTelemetry` ecosystem, my knowledge just stopped to the "instrumentation" part. I looked it and seems that this is aligned with my idea discussed in `Future Extensibility: Pluggable Architecture` this is exactly what the [OpenTelemetry Collector](https://opentelemetry.io/docs/) does. 

As how now, in our DS image, we have the possibility to [enable the `otel-agent`](https://github.com/debezium/container-images/blob/main/server/3.6/Dockerfile)  this should dynamically instrument the java code to  capture telemetry data at the “edges” of an app or service, such as inbound requests, outbound HTTP calls, database calls, and so on. ([java-agent doc](https://opentelemetry.io/docs/zero-code/java/agent/)). 

From my understanding this will not export current JMX metrics that instead requires the usage of [jmx-receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.150.0/receiver/jmxreceiver) that is deprecated and seems that the alternative is [jmx-scraper](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/jmx-scraper) that then should be use to scrape the JMX server getting also custom metrics and the push those to the open telemetry collector that can put the metrics to the Prometheus or whatever. 

Maybe we can use the current approach as a first working solution and then explore this more flexible one based on the needs? WDYT?

kmos: Well, in this case I’ll go directly with OpenTelemetry. I was aware of the extra work on the Debezium Server side, but I think it’s worth the effort: we should export our jmx metrics directly as otel metrics. I think it is also in the roadmap @jpechane

mfvitale: @kmos We could but I'm not confident to base the solution on the`jmx-scraper` that is in [alpha](https://github.com/open-telemetry/opentelemetry-java-contrib) state as of now. WDYT?

Naros: @mfvitale, have you thought about looking at the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) on OCP/K8? If I am not mistaken, Strimzi and OCP use this operator to collect metrics for both Connect & Connector CRD deployments, so it's already a key component in our commercial user base today. 

So I could imagine that if we built our UI around that stack, it would enable reuse when a community user transitions to a commercial user on OpenShift, or when a user chooses to pair the Platform with a commercial deployment Debezium alongside other Streams-based non-Debezium connectors.

In terms of OpenTelemetry, this again is probably something I wonder if we shouldn't think about modeling similar to Strimzi and how this is handled with `KafkaConnect` and `KafkaConnector` CRD deployments, for precisely the same reasons. 

What do either of you think?

Update: I guess I should have read more before commenting ;-).  I see you have.

kmos: > @kmos We could but I'm not confident to base the solution on the`jmx-scraper` that is in [alpha](https://github.com/open-telemetry/opentelemetry-java-contrib?rgh-link-date=2026-04-28T15%3A44%3A57Z) state as of now. WDYT?

I was thinking of exporting OpenTelemetry events directly, rather than scraping them from JMX. The idea would be to wrap our existing JMX implementation with an OpenTelemetry implementation.

jpechane: @kmos Using Prometheus also gives the user option to use Grafan himself if he has it deployed. Does OTel provide an equivalnt capability?

mfvitale: > @mfvitale, have you thought about looking at the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) on OCP/K8?

@Naros This is first option I proposed in `Option 1: Install Prometheus Operator (Default)` with `kube-prometheus-stack`

mfvitale: > > @kmos We could but I'm not confident to base the solution on the`jmx-scraper` that is in [alpha](https://github.com/open-telemetry/opentelemetry-java-contrib?rgh-link-date=2026-04-28T15%3A44%3A57Z) state as of now. WDYT?
> 
> I was thinking of exporting OpenTelemetry events directly, rather than scraping them from JMX. The idea would be to wrap our existing JMX implementation with an OpenTelemetry implementation.

This could be possible. So it means that we will have both support for JMX and native OTel. Then the OTel Collector should be configured with the [oltp receiver](https://github.com/open-telemetry/opentelemetry-collector/tree/v0.150.0/receiver/otlpreceiver) and we can still configure Prometheus as exporter. 

Given that in the proposed architecture this will just change the deployment part, we can think to have a first implementation with the proposed architecture: JMX exporter -> Prometheus (scrape) - REST query, and then we can add support to Debezium core for OTel native metrics so we can move (or give as alternative) to OTel Collector -> Prometheus (scrape) - REST query. 

@kmos The motivation to not start directly with OTel is that there are for sure some difference in how the metrics are named, instructed and also we need to verify that all metrics we have have a possible mapping with the OTel objects. Having a first working solution give us the calm to work on the OTel stuff without pressure. 

@jpechane @kmos We need to also understand where we want to go with `Micrometer monitoring using OpenTelemetry` that we have on the roadmap to avoid overlap. If I got it correctly this means use Micrometer as instrumentation library (it output Prometheus endpoint) and the export it via OTel. So given also the doubt expressed by @vjuranek in https://github.com/debezium/debezium/pull/6976#issuecomment-4341741276 maybe we can think to change the scope of  `Micrometer monitoring using OpenTelemetry` to `OpenTelemetry monitoring` so that we can go with my proposed solution with JMX now and when we have the OTel native metric switch or add the option to use OTel also in the Platform?

kmos: > @kmos Using Prometheus also gives the user option to use Grafan himself if he has it deployed. Does OTel provide an equivalnt capability?

Yes. Opentelemetry is totally compatible with [Grafana](https://grafana.com/oss/opentelemetry/). The main difference with prometheus is the approach (pull/push) and the kind of collections -> OTEL: metrics, logs, trace | Prometheus -> metrics. 

With otel you can be open to tracing: [you can trace](https://demo.jaegertracing.io/jaeger/trace/1885050a0942b5f086186bc28c262a6f) (and measure) for example a CDC event from the capturing to the transformation and so on...

> @kmos The motivation to not start directly with OTel is that there are for sure some difference in how the metrics are named, instructed and also we need to verify that all metrics we have have a possible mapping with the OTel objects. Having a first working solution give us the calm to work on the OTel stuff without pressure.

Totally reasonable

mfvitale: > With otel you can be open to tracing: [you can trace](https://demo.jaegertracing.io/jaeger/trace/1885050a0942b5f086186bc28c262a6f) (and measure) for example a CDC event from the capturing to the transformation and so on...

Regarding this we already have some [tracing](https://github.com/debezium/debezium/blob/c363a0b13cbffeb71225c970e1110b0ce672190b/debezium-connect-plugins/src/main/java/io/debezium/transforms/tracing/ActivateTracingSpan.java#L42) in place that we effectively [export via the otel agent](https://debezium.io/documentation/reference/stable/integrations/tracing.html).

vjuranek: > We need to also understand where we want to go with `Micrometer monitoring using OpenTelemetry` that we have on the roadmap to avoid overlap

+1 to have general discussion where we want to go with Debezium monitoring and tracing and what the long-term roadmap in this area will be

mfvitale: @kmos @vjuranek I have updated the proposal with the Otel Agent approach as discussed.

DDD-38.md
```diff
@@ -0,0 +1,898 @@
...
+#### Built-in Support
+
+The Debezium Server container image **already includes** the OpenTelemetry Java Agent. This will be restructured to align with the existing metrics infrastructure:
+
+**Current state:**
+- OpenTelemetry Java Agent is bundled in the Debezium Server image
+- No standardized configuration for Debezium-specific MBeans
+
+**Target state (to be implemented):**
+- Move the Otel Java Agent to the `debezium-server-dist` module under `src/main/resources/distro/lib_metrics/`
```
vjuranek, line 108: This is already done, right? I'd keep it tomake it visible that it was port of this DDD, but marked is as already done (possibly [x] or something)

DDD-38.md
```diff
@@ -0,0 +1,898 @@
...
+**Note:** The exact metric names depend on the JMX metric gatherer configuration. The names above follow OpenTelemetry naming conventions (snake_case). The mapping from Debezium JMX MBean attributes to Otel metric names is defined in the `debezium-jmx-config.yaml` file.
+
+### Component 5: Stage UI Integration
+
+**Responsibilities:**
+- Poll Conductor monitoring API at configurable intervals (5s, 10s, 30s, 1m, 5m)
+- Render time-series charts (line graphs, area charts)
+- Automatically use browser HTTP cache based on Cache-Control headers
+
+**Data flow:**
```
vjuranek, line 746: This is final end-to-end workflow. Maybe for better clarity I'd copy it also at the beginning next to Architecture diagram.

DDD-38.md
```diff
@@ -0,0 +1,898 @@
...
+**Evolution timeline:**
+
+**Phase 1 (Current):** OpenTelemetry Agent → Otel Collector → Prometheus
+- Zero code changes in Debezium Server
+- Agent collects JMX metrics via reflection
+- Metrics exported to Otel Collector via OTLP
+
+**Phase 2 (Future):** OpenTelemetry SDK → Otel Collector → Prometheus
+- Debezium Server integrates OpenTelemetry SDK directly
+- Native instrumentation replaces JMX MBeans
```
vjuranek: IIRC we agreed on creating  `debezium-metrics-jmx`  and `debezium-metrics-otel` modules (i.e. supporting both JMX and OTel), not replacing JMX by OTel.

vjuranek (commented):
Overall LGMT, just a few small comments.

DDD-38.md
mfvitale: Yes, I started to work while this is in discussion since that part is quite a stable part. i tracked all task with https://github.com/issues/assigned?reload=1&issue=debezium%7Cdbz%7C1894

DDD-38.md
mfvitale: Yes, what I intended here is that for the Platform we will then move to use native OTel. I'll rephrase a bit.

PR #43, opened 2026-05-13
add CONTRIBUTING.md to outline guidelines for writing and submitting Debezium Design Documents

vsantonastaso:


CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
+# Contributing a Debezium Design Document (DDD)
+
+This document explains how to write and submit a Debezium Design Document.
+A DDD describes a significant feature or architectural change from a design perspective, helping to shape the feature during development and allowing others to understand design decisions in retrospect.
+
+
+## Document Numbering
+
+Each DDD is assigned a sequential number: `DDD-1`, `DDD-3`, `DDD-7`, etc.
+Before creating a new document, check the existing documents and the [Contents list in README.md](README.md) to determine the next available number.
```
mfvitale: The next available number should be determine as follow: Open an issue on the `debezium-design-documents` repo with a short description and use the issue number as the DDD number.

mfvitale (changes requested):
Overall it LGTM. I left a comment to address

vsantonastaso:
Thanks @mfvitale. Comment addressed

mfvitale (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+2. Add an entry to the **Contents** list in [README.md](README.md):
+   ```markdown
+   * [DDD-<N>](DDD-<N>.md): <Short description>
+   ```
+3. Open a pull request against the `main` branch for community review.
+4. Iterate on feedback from maintainers and the community before merging.
+
+## Updating an Existing DDD
+
+Design documents are living artifacts. When implementation reveals that the design needs to change, update the DDD directly rather than creating a new one. Use pull requests for updates so that changes are reviewed and tracked.
```
kmos, line 110: I am not sure about this point: for example with ADR you create a brand new document also for an update.

kmos (approved):
only one comment but it may be something for the future

vjuranek (approved):
LGTM

CONTRIBUTING.md
vsantonastaso: Yes, but an ADR is a decision record, it contains a decision made at a specific point in time. If that decision changes in a different point in time, the correct approach is to create a new ADR, as you mentioned.
A DDD, instead, containes the design of a feature. It is tied to the feature itself and evolves as the feature evolves.
So, the key difference for me is that the ADR is time-specific, while DDD is feature-specific.
WDYT?

Naros: So I think there are two points that, at least to date, are at slight odds.

First, we generally have not updated DDDs after the initial implementation. Now, whether that's good or bad form, that's debatable, but it has not necessarily been our practice to date. I'm personally not opposed to doing it, and there are multiple reasons why it's good to do it, but it would be a "good" growing pain.

But I would prefer to avoid the ADR route. We used that approach in other projects in the past, and it's a complete nightmare for anyone joining the team midway through to track down and get caught up on a given feature, particularly if the organization of the ADRs is not well-defined, or if the ADR updates only focus on the "change" and not the full feature as a whole.

I think, given we're keeping these in version control, taking a hybrid approach where we have a DDD that we initially treat as an ADR, but then amend it as changes/improvements are made, seems reasonable. VCS gives us the granularity, and we could always keep a CHANGELOG section with a brief description of the why/what changed.

wdyt?

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Submission Process
+
+1. Create the `DDD-<N>.md` file and optional `DDD-<N>/` directory at the root of this repository.
+2. Add an entry to the **Contents** list in [README.md](README.md):
+   ```markdown
+   * [DDD-<N>](DDD-<N>.md): <Short description>
+   ```
+3. Open a pull request against the `main` branch for community review.
+4. Iterate on feedback from maintainers and the community before merging.
```
Naros: There should be a mention here that for any work related to a DDD to be accepted, it must adhere to the project governance rules, in the [Proposals](https://github.com/debezium/governance/blob/main/GOVERNANCE.md#proposals) section.

CONTRIBUTING.md
kmos: In my case, and in the context of updates, I used ADR as an example of a process in which you create a brand-new document. This makes the update clearer and more evident than a commit that does not take into account the motivation behind the document. Timing is important in ADR because changes happen very quickly, and sometimes the document evolves together with the code, which is not our case.

I was thinking more about the process used for [KPI](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals), but I couldn’t understand from their documentation whether they update an existing KIP or create a new one. Maybe @mfvitale knows. 

In any case, it is already approved since it’s something we can postpone for now.

Naros: I'm pretty sure their KIP process follows more the ADR approach you described. To be fair, my disdain for ADR is mainly because the process we used involved storing these on google docs :/

kmos: > But I would prefer to avoid the ADR route. We used that approach in other projects in the past, and it's a complete nightmare for anyone joining the team midway through to track down and get caught up on a given feature, particularly if the organization of the ADRs is not well-defined, or if the ADR updates only focus on the "change" and not the full feature as a whole.

Sorry, I mentioned ADR but my point is only how to manage the update. I don't see any room to apply that process

CONTRIBUTING.md
vsantonastaso: Done. Thanks @Naros

Naros (approved):

PR #45, opened 2026-05-17
DDD-41: Host-Based Pipeline Deployment for the Debezium Platform GSoC'26

d1vyanshu-kumar:
Currently, the Debezium Platform supports pipeline deployments exclusively through the Kubernetes-based Debezium Operator. This DDD proposes a first-class, parallel alternative to support host-based deployments (bare-metal servers, cloud VMs, and on-premise machines) that do not run Kubernetes.

By implementing a host-based path utilising SSH for initial provisioning and a lightweight remote agent for lifecycle management, the Debezium Platform becomes environment-agnostic.


DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+This proposal introduces a host-based deployment path as a first-class, parallel alternative to the existing Kubernetes path. By implementing a second `EnvironmentController` that deploys Debezium Server containers on remote hosts via SSH and a lightweight host-side agent, the Platform becomes environment-agnostic.
+
+## Proposed Changes
+
+### 1. High-Level Architecture
+
+The existing deployment flow is:
+
+```
+Stage UI → REST API → PipelineService → Outbox Event → Watcher (Debezium Engine)
```
mfvitale: Maybe will be good to have an architecture diagram. You can use any tool you like or [excalidraw](https://excalidraw.com/)

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+```
+
+Note: `fetch = FetchType.LAZY` on `@ManyToOne` to avoid N+1 queries in the status poller. `serverPort` field added for per-pipeline port allocation.
+
+### 5. SSH Layer
+
+Uses Apache MINA SSHD (`org.apache.sshd:sshd-core`, `org.apache.sshd:sshd-sftp`):
+
+```java
+package io.debezium.platform.environment.host.ssh;
```
mfvitale: It could be good to have an high level description/diagram on how the things will works, avoiding putting to much details in this phase.

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+    HostTargetEntity.ProvisioningStatus status,
+    List<CheckResult> checkResults,
+    Instant timestamp
+) {}
+```
+
+| Check | SSH Command | Remediation | Notes |
+|---|---|---|---|
+| `ConnectivityCheck` | `echo debezium-health-check` | None (fail-fast) | |
+| `DockerInstalledCheck` | `docker version --format '{{.Server.Version}}'` | `apt-get install -y docker.io` or `yum install -y docker` | Detects OS via `/etc/os-release` |
```
mfvitale: Could you investigate the creation of an [Ansible Playbook](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_intro.html) for the host setup??

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+Restart=always
+RestartSec=5
+
+[Install]
+WantedBy=multi-user.target
+```
+5. Runs `sudo systemctl daemon-reload && sudo systemctl enable debezium-agent && sudo systemctl start debezium-agent`
+6. Verifies agent is responding: `curl -s -o /dev/null -w '%{http_code}' http://localhost:{agentPort}/api/status`
+
+#### 6.3 Provisioning Service
```
mfvitale: Too much details for this phase. Could you describe how the provisioning will work at high level?

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+        });
+    }
+}
+```
+
+Note: `DockerPermissionsCheck` remediation requires SSH session reconnect because `usermod -aG docker $USER` only takes effect in new login sessions. The explicit `try/finally` ensures whichever session is current (original or reconnected) is always closed.
+
+### 7. Host Pipeline Mapper
+
+**`HostPipelineMapper`** — analogous to `PipelineMapper` in `io.debezium.platform.environment.operator`. Both consume a `PipelineFlat`; `PipelineMapper` outputs a `DebeziumServer` CRD, `HostPipelineMapper` outputs a flat `application.properties` string.
```
mfvitale: Same as above. Too much details for this doc.

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+    }
+
+    @Override
+    public void undeploy(Long id) {
+        // No-op. See deploy() rationale above.
+    }
+}
+```
+
+**Why no-op?** In Kubernetes, `OperatorVaultController` deploys secrets as Kubernetes Secret objects that Pods reference. On bare-metal hosts, there's no equivalent secret store. Pipeline credentials are resolved at config-generation time by `HostPipelineMapper` and written directly into `application.properties`. SSH credentials for host access are stored in the existing `VaultEntity` and resolved by `HostProvisioningService` when opening SSH sessions. The `OperatorVaultController` in the existing codebase is ALSO effectively empty (methods have empty bodies) — so this is consistent.
```
mfvitale: Regarding the SSH key pairs I recall that we [discussed](https://github.com/debezium/debezium-design-documents/pull/30/changes#r3012077375) a possible solution using secrets/volumes. 

At this time I think we can investigate the approach to use the volumes, since this could work when Platform is deployed  on K8s and docker deployment. 

Could you please explore this and see how this integrates with `Apache MINA SSHD`?

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+```
+
+**Key points:**
+1. `docker rm -f ... 2>/dev/null || true` — prevents exit code 1 on first deploy
+2. Uses `request.serverPort()` instead of hardcoded `8080` — avoids port conflicts
+3. Bearer token auth via `BearerTokenFilter` `@Provider` — secures the agent API
+
+### 9. Host Agent Client (Conductor-side)
+
+**Cannot use MicroProfile `@RegisterRestClient`** because each host has a different URL. Must use programmatic HTTP client:
```
mfvitale: Try to have a look to [dynamic-base-urls](https://quarkus.io/guides/rest-client#dynamic-base-urls)

DDD-41.md
```diff
@@ -0,0 +1,1158 @@
...
+
+**`Hosts.tsx`** — List page showing all registered hosts with provisioning status badges (READY=green, PROVISIONING=blue spinner, FAILED=red).
+
+**`CreateHost.tsx`** — Registration form with fields: name, hostname, port (default 22), username, auth type (dropdown: SSH_KEY/PASSWORD), credential (Vault selector), agent port (default 8090).
+
+**`HostDetails.tsx`** — Detail page showing host info, provisioning status with check results, list of deployed pipelines on this host, and re-provision button.
+
+#### 12.4 Pipeline Designer Modification
+
+In the pipeline creation/edit flow, add a "Target Environment" step where the user can choose:
```
mfvitale: In general the Platform goal is to let user focusing on the pipeline designing and not on how things are then deployed. 

So I could image, at least in this initial phase, that users select the deployment model at deploy time. So you can use the Platform to deploy on k8s or on bare metal. @jpechane , @Naros WDYT on this?

Also from the UI perspective the section related to the configuration of the hosts will be only displayed it the deployment mode is bare metal. 

Said that, we should think how the automatic deployment works in bare metal when there are multiple hosts: round robin, by machine load?

mfvitale:
@d1vyanshu-kumar thanks! I added some comments. 

In general in this kind of documents, you should treat the argument in high level with some details where needed to facilitate the discussion. 

Key important point to define are: 

- How we manage the SSH key pair
- What deployment model we want to achieve.

DDD-41.md
d1vyanshu-kumar: Thanks for the suggestion! I've added a comprehensive architecture diagram (see Section 1).

DDD-41.md
d1vyanshu-kumar: Thanks! I've stripped away the unnecessary implementation details and refocused the document on the core architecture. Please let me know if there are still areas that need improvement or further refining.

DDD-41.md
d1vyanshu-kumar: Thanks for the suggestion. I looked into Ansible, and using a playbook is definitely the right call here. It allows us to handle the host setup declaratively rather than writing a bunch of imperative logic to manage raw SSH connections.

I've modified Section 6 to replace the Java SSH provisioning code with an Ansible Playbook approach. The playbook handles Python bootstrapping, Docker installation, daemon configuration, Docker group permissions, Agent deployment, health verification, and Debezium Server image pre-pulling—all as idempotent tasks.

This eliminates the need for custom Java SSH classes (like MINA SSHD). The Conductor will simply invoke the playbook via ProcessBuilder, passing the SSH key path using the --private-key flag.

DDD-41.md
d1vyanshu-kumar: done  👍

DDD-41.md
d1vyanshu-kumar: I've updated the Agent Client design (Section 9) to use Quarkus Dynamic REST Clients.

DDD-41.md
d1vyanshu-kumar: I've redesigned the SSH key management (Section 5). Private keys are now stored as files, not in the database. In Kubernetes deployments, the key is stored as a K8s Secret and mounted as a volume into the Conductor pod.

DDD-41.md
d1vyanshu-kumar: done!

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+A new Flyway migration creates the `host_target` and `host_deployment` tables with appropriate sequences, foreign keys to `pipeline`, and indexes.
+
+### 5. SSH Key Pair Management
+
+SSH private keys for connecting to remote hosts are managed using **volume mounts**, not stored in the database.
+
+#### 5.1 How It Works
+
+1. The user creates a file (or Kubernetes Secret) containing the SSH private key.
+2. The file is mounted into the Conductor pod at a known path (e.g., `/etc/debezium/ssh-keys/`).
```
mfvitale: i think that we can leverage the automatism of [mina-sshd](https://github.com/apache/mina-sshd/blob/master/docs/client-setup.md#clientidentityloaderkeypairprovider) mounting in `the user's ~/.ssh folder` and I think it should then automatically pick the correct private key based on the host (not 100% sure). Maybe a quick test is worth.

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+|---|---|
+| **Kubernetes** | Create a K8s Secret with the private key → mount as a volume into the Conductor pod |
+| **Docker Compose** | Bind mount a host directory: `-v /path/to/keys:/etc/debezium/ssh-keys/` |
+| **Development** | Place key files in a local directory referenced by config |
+
+This approach has several advantages:
+- Key material never touches the database
+- Kubernetes handles encryption at rest and RBAC natively
+- Works identically for both K8s and Docker deployments of the Platform itself
+- No SSH libraries needed in Java — Ansible reads the key file natively via `--private-key`
```
mfvitale: It is not clear to me if SSH will be used to first connect to the machine and the use Ansible playbook to setup the machine or Ansible is used to to also the initial connection. 

I'm not an Ansible expert so could you plaese clarify?

d1vyanshu-kumar: Hi @mfvitale, to clarify: Ansible handles the initial connection entirely on its own. Java doesn't open any SSH channels to the remote machine.
 
As mentioned in the [Ansible documentation](https://docs.ansible.com/projects/ansible/latest/inventory_guide/connection_details.html#controlpersist-and-paramiko), Ansible natively uses standard OpenSSH to handle network connections. We just pass the key file path natively using the `--private-key` command line option when we trigger the playbook.

 Essentially, Java's only job is to trigger the `ansible-playbook` CLI process - Ansible then takes over to handle 100% of the authentication, connection, and machine setup.

DDD-41.md
d1vyanshu-kumar: That's a great point about automatically picking up keys from the ~/.ssh folder!

However, because we have decided to use Ansible to handle the host provisioning (as discussed in Section 6), we actually don't need the mina-sshd Java library at all anymore.

The great news is that Ansible (since it uses native OpenSSH) already does exactly what you are suggesting! If we mount the keys into the Conductor's ~/.ssh/ folder, Ansible will automatically detect and use the correct private key for the host without us needing to write any Java code to map it.

I proposed passing the path explicitly via --private-key just to give users flexibility if they mount the key in a custom location, but we can definitely default to the standard ~/.ssh/ automatic mounting behavior!"

mfvitale: > I proposed passing the path explicitly via --private-key just to give users flexibility if they mount the key in a custom location, but we can definitely default to the standard ~/.ssh/ automatic mounting behavior!"

Good point. 

So effectively we have two options:

1. Let users provide the ssh config file and mount it to through a volume and the detect that file and use it to configure the declared hosts with Ansible

So effectively users provides 

```text
Host targaryen
    HostName 192.168.1.10
    User daenerys
    Port 7654
    IdentityFile ~/.ssh/targaryen.key
Host tyrell
    HostName 192.168.1.20
    User margaery
    Port 7654
    IdentityFile ~/.ssh/tyrell.key
```

and then the provision service will use Ansible to setup all hosts with 

``` shell
ansible all -i inventory.ini -m ping`
```

2. Use the approach where users defines the Hosts on the UI, referencing the key file for the specific host but still a volume will be used but in future we could improve adding the possibility to upload the key file and store them in the volume. 

Option 1 seems for me much easy from the implementation point of view but there could be some sort of friction from the user level since the hosts must be added at deploy time. On the other end host are normally configured by some sysadmin/ops guys in any case. We can thing to an automatic mechanism that watches if any changes on the config file so understand if an host has been removed or added or IP changes. 

This will remove the needs to have the host table and UI.

Option 2 could be a possible evolution if there is more demand on this area. @jpechane @Naros @kmos @vsantonastaso I would like to have your thoughts on this. 

Option 1 also seems more aligned with our claim that on the Platform users focus on the pipeline design rather then deployment.

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+
+### 3. Core Abstractions — Routing
+
+#### 3.1 The CDI Challenge
+
+The `PipelineConsumer` and `VaultConsumer` each receive an `EnvironmentController` via constructor injection. Currently, `OperatorEnvironmentController` is the only `@ApplicationScoped` bean implementing this interface. Adding a second implementation would cause CDI ambiguity.
+
+#### 3.2 Solution: RoutingEnvironmentController
+
+A new `RoutingEnvironmentController` becomes the single `@ApplicationScoped` `EnvironmentController` bean. It wraps both `OperatorEnvironmentController` and `HostEnvironmentController` (both changed to `@Dependent`) and delegates based on whether a `HostDeploymentEntity` exists for the given pipeline.
```
mfvitale: Since we have defined that the Platform will be deployed in one of the two mode, I thin the router is no more needed since you will use `OperatorEnvironmentController` or `HostEnvironmentController` not both. 

What is still not clear to me is who and how decides where the target host for a pipeline? 

I would avoid a manual mapping from the UI.

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+The lifecycle of a host-deployed pipeline mirrors the Kubernetes path conceptually:
+
+| Operation | Kubernetes Path | Host Path |
+|---|---|---|
+| **Deploy** | `serverSideApply()` CRD → Operator creates Pod | HTTP POST to remote host → Docker run |
+| **Undeploy** | Delete CRD → Operator removes Pod | HTTP POST to remote host → Docker rm |
+| **Stop** | Set `stopped=true` on CRD | HTTP POST to remote host → Docker stop |
+| **Start** | Set `stopped=false` on CRD | HTTP POST to remote host → Docker start |
+| **Logs** | K8s Pod log API | HTTP GET from remote host → Docker logs |
+| **Signals** | REST call to Debezium Server K8s Service | HTTP call proxied through remote host |
```
ismailsimsek: is this endpoint going to be added to Debezium Server? where signals are posted to the Debezium Server directly. that will be great addition.

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+
+The Platform's goal is to let users focus on pipeline design, not on how things are deployed. Therefore:
+
+- **Pipeline design** remains unchanged — users configure source, destination, transforms as today
+- **Deployment model selection** happens at **deploy time**, not at design time
+- When deploying, the user selects either:
+  - **Kubernetes** (default, existing behavior)
+  - **Bare Metal** → selects a registered host with `READY` provisioning status
+
+The host management section in the UI is only displayed when the deployment mode is bare-metal.
```
ismailsimsek: i know this is less ideal, but would it be possible to deploy everything to the same host of the Platform? instead of remote using current localhost? where Platform, Agent and Debezium Server works on same host.

DDD-41.md
```diff
@@ -0,0 +1,344 @@
...
+
+### 2. Deployment Model Selection
+
+The Platform's goal is to let users focus on pipeline design, not on how things are deployed. Therefore:
+
+- **Pipeline design** remains unchanged — users configure source, destination, transforms as today
+- **Deployment model selection** happens at **deploy time**, not at design time
+- When deploying, the user selects either:
+  - **Kubernetes** (default, existing behavior)
+  - **Bare Metal** → selects a registered host with `READY` provisioning status
```
ismailsimsek: would it be possible to run multiple pipelines inside single Bare-metal host?

ismailsimsek (commented):
Thank you @d1vyanshu-kumar for working on this, great feature. Looks great to me, left few comments/questions.

DDD-41.md
mfvitale: @ismailsimsek The [endpoint to send signals](https://github.com/debezium/debezium-server/blob/13edabf731657729c93e15fde9dbb7dee641c4d3/debezium-server-core/src/main/java/io/debezium/server/rest/signal/SignalResource.java#L28) to DS already exist thought it is not documented since they was initially meant to be used by the Platform.

DDD-41.md
mfvitale: Since we want to leverage on Ansible, this should be perfectly possible. 

Just a clarification. We are focusing on having the Platform deployed on K8s and then deploy pipeline on hosts, but I could image that running the Platform with a `docker-compose` on a host would not be so difficult. So effectively this could be a possibility.

DDD-41.md
mfvitale: Yes, on deployment you define a set of hosts to use, there will be a strategy that define where a pipeline is deployed, like round robin or by available resources. 

You can still define a single host and all pipelines will be deployed there.

DDD-41.md
d1vyanshu-kumar: > i know this is less ideal, but would it be possible to deploy everything to the same host of the Platform? instead of remote using current localhost? where Platform, Agent and Debezium Server works on same host.

Yes, absolutely! Because the architecture is completely driven by the ~/.ssh/config file and Ansible ([section-3](https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/2f320972d85d331b17eff3cb9ccbbbfb2d13f98c/DDD-41.md#3-host-discovery-via-ssh-config-file-watcher), [section-6](https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/2f320972d85d331b17eff3cb9ccbbbfb2d13f98c/DDD-41.md#6-host-provisioning-via-ansible-playbook)), deploying everything to the same host is natively supported out of the box.

You would just add a localhost entry to the SSH config file. The Conductor will discover it, Ansible will provision the Agent locally, and pipelines will deploy there. The system doesn't differentiate between a remote IP and 127.0.0.1, so this use case works perfectly without any architectural changes!

DDD-41.md
d1vyanshu-kumar: @ismailsimsek Yes, absolutely! Just as @mfvitale mentioned, this is fully supported and is actually a core part of the design.

We have detailed this in [Section 7](https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/2f320972d85d331b17eff3cb9ccbbbfb2d13f98c/DDD-41.md#7-automatic-pipeline-scheduling-concurrency-safe) (Automatic Pipeline Scheduling). When a pipeline is deployed, the Conductor uses a RoundRobinStrategy to automatically select the host with the fewest active pipelines. If you only have one host registered in your SSH config, all pipelines will naturally be deployed there.

To ensure multiple Debezium Server containers can run on the same bare-metal host without crashing into each other, the Conductor automatically allocates a unique, conflict-free port (MAX(serverPort) + 1) for each pipeline, as detailed in Section 9.

So yes, running multiple pipelines on a single host works seamlessly out of the box!

d1vyanshu-kumar:
Hi @ismailsimsek,

Thanks so much for taking the time to review my design document! Sorry for the delay in getting back to you; I am currently in the middle of my semester exams.

I have updated the architecture based on the recent discussions (including adopting the SSH config for host discovery and refining the scheduling strategy). I believe these changes provide a much clearer picture of the implementation.

Could you please take another look when you have a chance? 
Thank you again for your time!

mfvitale:
@d1vyanshu-kumar LGTM. Great work.

mfvitale (approved):

DDD-41.md
kmos: IMHO, Option 1 is the way for now

DDD-41.md
ismailsimsek: fantastic, thank you

ismailsimsek (approved):

kmos (approved):

PR #47, opened 2026-05-25
DDD-50 for PyDebeziumAI real-time CDC integration

KMohnishM:
Hi @jpechane  and @vjuranek ,

I have drafted the Debezium Design Document (DDD-50) for **PyDebeziumAI** outlining the library's architecture and design details.

This PR adds:
- **DDD-50.md**: The complete design proposal describing the 5-layer pipeline (Ingestion, Canonical, Transformation, Sync, and Adapters), configuration examples, consistency guarantees, and mitigations.
- **DDD-50/architecture.png**: Generated system architecture diagram.
- **DDD-50/architecture.mermaid**: The Mermaid source code of the diagram for future editing.

I would love to get your thoughts and feedback on this! Thanks!


KMohnishM:
Also I have attached a generated image , would change to [excalidraw](https://excalidraw.com/) if thats preffered .

DDD-42.md
```diff
@@ -0,0 +1,294 @@
+# DDD-42: PyDebeziumAI - Real-Time CDC Integration for LangChain & LangGraph
+
+This document describes the design and architecture of **PyDebeziumAI**, a Python library that enables real-time Retrieval-Augmented Generation (RAG) synchronization by streaming database change events captured by Debezium into vector stores.
+
+## Motivation
+
+Modern Generative AI and Large Language Model (LLM) applications rely on Retrieval-Augmented Generation (RAG) to ground model responses in dynamic domain data. Relational databases store the source of truth for these applications, but traditional methods for syncing database state to vector search indexes have significant limitations:
+
+1. **Periodic Batch Refreshes / Polling**: Querying the database periodically is slow, resource-intensive, and introduces sync lag. During the lag window, the LLM retrieves stale context, leading to hallucinations.
```
vjuranek: Not sure if it's correct to call this hallucination when the result is based on the old data. IMHO something like "output based on outdated data or hallucination when no data is available" would be more appropriate.

DDD-42.md
```diff
@@ -0,0 +1,294 @@
...
+- **Metadata Sanitizer**: Coerces non-standard data types into vector-compatible primitives:
+  - `Decimal` ──► `float`
+  - `bytes` / `bytearray` ──► `hex-encoded string`
+  - `datetime` / `date` / `time` ──► `ISO 8601 string`
+  - `dict` / `list` ──► `string representation`
+
+#### 2.4 Synchronization Layer (`SyncManager`)
+Executes vector store commands according to CDC transaction operation types:
+- **`c` (Create) and `r` (Snapshot)**: Performs an `adapter.upsert()`.
+- **`u` (Update)**: Calls `adapter.delete()` on the document ID, followed by `adapter.upsert()`. This ensures that if metadata changes significantly, stale embeddings are completely purged.
```
vjuranek: If we have upsert operation, why do we need to call delete beforehand?

DDD-42.md
```diff
@@ -0,0 +1,294 @@
...
+- **Out-of-Order Handling**: Deduplication and synchronization are anchored on the unique primary key mapping strategy (`TablePkIdStrategy`). Because updates are idempotent upserts on the vector store, out-of-order deliveries resolve to the latest database state.
+- **Atomic Updates**: To ensure that changing a database row does not leave dangling metadata keys or duplicate embeddings in the vector index, updates execute a sequential delete-then-upsert pipeline on the vector store adapter.
+- **Delete-then-Upsert Failure Recovery**: If the subsequent upsert operation fails after a delete succeeds, the document temporarily disappears from the vector database index. To close this consistency gap, the failed event is redirected to the `DeadLetterQueue` (DLQ), retaining the message sequence context. Replay workers read from the DLQ to retry the upsert, thereby restoring consistency without permanent data loss.
```
vjuranek: Who are replay workers synchronized with the main update worker?

vjuranek (commented):
Nice work @KMohnishM ! I left a few questions.

DDD-42.md
KMohnishM: Good point. I have updated the document to refine this

DDD-42.md
KMohnishM: To prevent race conditions where a replay worker processes a delayed retry event from the DLQ *after* the main update worker has already applied a newer write/update to the same record, we pause/block processing of subsequent updates for that specific primary key while it is undergoing retries in the DLQ. 
 
 This guarantees that updates are always processed sequentially per logical row. I have updated the design document to detail this synchronization policy.

DDD-42.md
KMohnishM: The reason is that in many vector databases (e.g., Chroma or Milvus), a standard `upsert` or `update` operation merges the new metadata dict with the existing metadata dict instead of fully replacing it. 
 
 If a column in the relational database is set to NULL (deleted) or metadata keys are removed, a simple upsert will leave the old keys as 'dangling metadata' in the vector store. Performing a `delete` followed by `upsert` guarantees that the vector store's representation strictly matches the new database row state without retaining stale metadata keys. I have added this explanation to the design document.

DDD-42.md
```diff
@@ -0,0 +1,294 @@
+# DDD-42: PyDebeziumAI - Real-Time CDC Integration for LangChain & LangGraph
+
+This document describes the design and architecture of **PyDebeziumAI**, a Python library that enables real-time Retrieval-Augmented Generation (RAG) synchronization by streaming database change events captured by Debezium into vector stores.
+
+## Motivation
+
+Modern Generative AI and Large Language Model (LLM) applications rely on Retrieval-Augmented Generation (RAG) to ground model responses in dynamic domain data. Relational databases store the source of truth for these applications, but traditional methods for syncing database state to vector search indexes have significant limitations:
+
+1. **Periodic Batch Refreshes / Polling**: Querying the database periodically is slow, resource-intensive, and introduces sync lag. During the lag window, the LLM retrieves stale context, leading to outputs based on outdated data, or hallucinations if the requested entity was deleted in the database but still exists in the vector store.
```
vjuranek: I'd say that this is still result based on the obsolete data and hallucination is the case when there are no data at all, AI tries to figure out result and due to lack of the data "hallucinate" the result. So if you want o add some example, it would be that e.g. new product was added into the database, but the data wasn't propagated to the vector store and there is no information about the new product, if the customer asks AI agent about new product, it may hallucinate the reply, which can be completely misleading.

vjuranek (commented):
Thanks for the updates @KMohnishM . I left one more comment again about hallucination, but this is just a detail. Let's wait for @jpechane if he heas any comments.

DDD-42.md
KMohnishM: Yeah got it , i will update this

kmos (changes requested):
@KMohnishM please change the DDD number accordingly to the [issue](https://github.com/debezium/debezium-design-documents/issues/50): DDD-50

kmos (approved):

Issue #42, opened 2026-05-11
Milvus Source Connector


kmos:
# Description

This project proposes the development of a Source Connector for Milvus, a high-performance vector database widely used in AI/ML-driven applications such as semantic search, recommendation systems, Retrieval-Augmented Generation (RAG), and multimodal search.

The connector will enable streaming and ingestion of vectorized data and metadata from Milvus into downstream systems, facilitating integration with data pipelines, analytics platforms, or event-driven architectures. It will continuously monitor changes in Milvus collections (e.g., inserts, updates, deletions) and convert them into a standardized event format.

cc @kartikangiras

kmos:
@kartikangiras can I assign it to you?

kartikangiras:
@kmos yeah sure.

PR #48, opened 2026-05-28
DDD-42 Debezium Source Connector for Milvus

kartikangiras:
Milvus is a leading vector database, but Debezium currently only supports Milvus as a sink, not as a source. This means changes within Milvus (inserts, deletes, schema changes) are not visible to the broader data ecosystem.

This proposal introduces a Debezium Source Connector for Milvus, initially targeting Milvus 2.5 with Kafka MQ, but architected for easy extension to Milvus 2.6+ (Woodpecker StreamingNode). 
The connector will:

- Capture all DML (inserts, deletes) and relevant DDL (schema changes) events from Milvus.
- Use Kafka MQ as the transport for 2.5, with a clear seam for supporting Woodpecker WAL in 2.6+.
- Leverage etcd for metadata, schema, and checkpoint alignment.
- Ensure strict ordering and correctness using TSO and timetick semantics.
- Integrate with Debezium’s schema and offset management for reliable, restartable streaming.
- Reuse Debezium’s core framework for rapid development and robust operation.
This approach closes a critical gap, enabling Milvus to participate fully in change data capture pipelines and hybrid data architectures.

Issue #42


DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+
+**Actual deserialization path**:
+1. Parse Kafka record value as `MsgPack` first.
+2. If `MsgPack` parsing fails, retry as single message payload.
+3. Extract `MsgType` from `MsgBase`.
+4. Convert to one of: `InsertEvent`, `DeleteEvent`, `CreateCollectionEvent`, `DropCollectionEvent`, `TimeTickEvent`.
+5. For `InsertRequest`, run column-to-row pivot (`num_rows` aligned across all fields).
+6. On unknown message type or malformed payload, throw `DebeziumException` and fail task (no silent skip).
+
+**Note:** The Kafka topic stores `MsgPack` or individual serialized messages depending on the Milvus version's `msgstream` implementation. Validate the exact wire format by inspecting a live cluster before writing the deserializer.
```
Naros: The note here seems to indicate that we should determine the deserialization path by doing a discovery pass on the Milvus cluster, but the steps seem to indicate more of a try-fallback style, which reads as contradictory imo.

So the goal here is to discover the deserialization path we should take by looking at the server, and then read the topics? If so, is there any potential case where across a cluster update the deserialization of older events could have a different format than newer ones?

DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+  void assignAndSeek(Map<TopicPartition, Long> offsets);
+  List<RawMilvusMessage> poll(Duration timeout);
+  void close();
+}
+```
+
+**KafkaMilvusMessageConsumer implementation details:**
+- Uses `consumer.assign()` only.
+- Calls `seek(tp, offset+1)` on warm restart.
+- Uses bounded poll timeout and propagates retriable Kafka exceptions.
```
Naros, line 398: This needs to be clearly documented in the code because this bypasses consumer-group semantics. While deliberate, future maintainers may find this surprising or not understand the justification.

DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+1. Bytes `[0:8]` = magic number / header
+2. Remaining bytes = serialized `MsgBase` + message-specific proto
+
+In practice, the message stream uses `github.com/milvus-io/milvus-proto` and the `MarshalType` determines whether the header is included.
+
+**Actual deserialization path**:
+1. Parse Kafka record value as `MsgPack` first.
+2. If `MsgPack` parsing fails, retry as single message payload.
+3. Extract `MsgType` from `MsgBase`.
+4. Convert to one of: `InsertEvent`, `DeleteEvent`, `CreateCollectionEvent`, `DropCollectionEvent`, `TimeTickEvent`.
```
Naros: One thing we should discuss is whether it makes sense to have `upsert` operations replicated by Debezium in the same way that Milvus writes them to the topic (`delete` followed by `insert`) or if we should should correlate these into a true `update` iff (if and only if) the primary keys are identical.

There are pros & cons to either approach and could depend on target.

DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+
+**Buffer size limit:**
+- `maxBufferedEvents` (count)
+- `maxBufferedBytes` (approx payload bytes)
+- crossing either threshold triggers backpressure path
+
+When `MilvusBufferFullException` is thrown, the streaming source should:
+1. Stop calling `consumer.poll()` temporarily
+2. Force-flush with emergency watermark after `timetickStallTimeout`
+3. Resume normal polling
```
Naros: Is there a potential for a late message from a stalled vchannel with a lower TSO to arrive after the flush here, and this creates an out-of-order emission?

DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+- Emit one Debezium record per reconstructed row.
+
+Milvus sends all values for field F as a contiguous array inside `FieldData`. Row `i` is reconstructed by taking index `i` from each field's array. The connector must pivot this into `num_rows` individual row maps before dispatching Debezium events.
+
+- [msg.proto — full message definitions](https://github.com/milvus-io/milvus-proto/blob/master/proto/msg.proto)
+- [schema.proto — FieldData, CollectionSchema](https://github.com/milvus-io/milvus-proto/blob/master/proto/schema.proto)
+- [Milvus msgstream Go source — serialization](https://github.com/milvus-io/milvus/blob/master/pkg/mq/msgstream/msgstream.go)
+- [Milvus internal column-to-row handling](https://github.com/milvus-io/milvus/blob/master/internal/storage/utils.go#L540-L582)
+
+### 3.5 Etcd Layout
```
Naros: Do you intend to use the Milvus API to read from etcd or are we going to etcd directly? The latter is a risk, and something that is deemed internal to Milvus, and is subject to change.

DDD-42.md
```diff
@@ -0,0 +1,639 @@
...
+
+2. **`WoodpeckerOffset`** — a new offset type storing `{vchannel, messageId}` pairs instead of Kafka topic/partition/offset.
+
+3. **`EtcdMetadataReader` changes** — the etcd key layout for channel checkpoint changes in 2.6 (Woodpecker stores its own checkpoints). The reader must be updated to read from the correct prefix.
+
+4. **No changes required to:** `TimetickOrderingEngine`, `MilvusProtoDeserializer`, `MilvusSchema`, `MilvusEventConverter`, `VectorFieldSerializer`, `MilvusStreamingChangeEventSource`.
+
+DDL via the Woodpecker WAL is more complex than etcd watch — DDL entries in the WAL require more parsing. A hybrid approach (etcd watch for DDL, Woodpecker for DML) as described in the Proposal A document is the correct design for 2.6.
+
+## 12. Resources and References
```
Naros: I don't see any section here that discusses metrics and/or observability. We should have a section that outlines at least the basic MXBean properties for snapshot & streaming, and how those will be mapped to runtime sources/calculations. This way we can identify if there are potential gaps where we cannot supply a specific metric or where there may be a need for additional connector-specific metrics.

Naros (commented):
Hi @kartikangiras I'll do a more thorough review, but this is just something to give you some food for thought.

DDD-42.md
kartikangiras: AS discussed with you in the meeting, upsert operations replication (d followed by i) is a better approach and will be used for the implementation.

DDD-42.md
kartikangiras: I have updated it for the discovery based approach.
I don't think so there might be a case where a conflict of formats in respect to deserialisation of events could occur because it would then eventually break up milvus-cdc and it is a milvus internal thing.

DDD-42.md
kartikangiras: Yeah, it is a possibility. Using the force-flush could be the forced or the last recovery path and providing some safeguards around this could make it work.

DDD-42.md
kartikangiras: Milvus API would be preferred to read from etcd, but adding direct etcd reading could be used as a fallback method in some cases where Milvus API does not work. wdyt?

DDD-42.md
kartikangiras: I have added the section which discusses all these metrics and gaps in detail.

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+- [Milvus time synchronization design](https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20211215-milvus_timesync.md)
+
+### 3.4 Message Types and Protobuf Wire Format
+
+All Milvus MQ messages share a common envelope. The connector must handle these types:
+
+| MsgType enum | Proto message | Connector action |
+|---|---|---|
+| `Insert` (1) | `InsertRequest` | Pivot columns→rows, emit `op=c` per row |
+| `Delete` (2) | `DeleteRequest` | Emit `op=d` per primary key |
```
Naros: This is one area that is likely going to require some special documentation. Given that the `Delete` event does not carry prior state, deletes from Milvus are effectively Tombstones by default and nothing more.

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+
+We should also verify the connector in Debezium Server mode, not only in Kafka Connect mode.
+
+Minimal verification checklist:
+
+1. Boot Debezium Server with `debezium.source.connector.class` pointing to `MilvusConnector`.
+2. Verify snapshot records reach configured sink.
+3. Insert additional rows after startup and verify streaming records reach sink.
+4. Restart Debezium Server and verify resume happens from stored offsets.
+5. Force an offset-expired scenario and verify `snapshot.mode` behavior (`initial` vs `never`).
```
Naros, line 765: While i would agree these are minimal, we should make sure the design we go without is potentially capable of easily supporting `when_needed`, `no_data`, and `recovery`.

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+
+DDL events (CreateCollection, DropCollection) arrive in the same Kafka stream as DML events. They carry TSOs and participate in the timetick ordering. This guarantees schema-before-data ordering. If `CreateCollectionEvent` at TSO=T is flushed, inserts for that collection with TSO>T are flushed later.
+
+**Schema change processing in the flush path:**
+1. `CreateCollectionEvent`: build table schema then emit schema change.
+2. `DropCollectionEvent`: emit schema drop event then remove from schema cache.
+3. DML event dispatch is blocked until schema state for that collection is valid.
+
+**Schema evolution between restarts:**  
+On restart, `MilvusMetadataClient` re-reads current schemas from Milvus gRPC metadata APIs. The connector does not rely on the schema history topic to reconstruct the current schema. The schema history topic is written for downstream consumers only.
```
Naros: When you re-read the schemas, how do you guarantee that the schema you read is specifically the schema that should be used for that specific collection's offset position? 

This is one area I believe needs some thinking because in relational connectors, this is solved by tracking schema changes in a history topic, and recovering the history topic upon restart guarantees that the schema we have is what is expected for the next event read. So think of it like reading the metadata during `initial`, `no_data`, or `recovery` snapshot modes, but otherwise its always recovered from the topic and not the database.

Could this approach work, at least when Milvus is operating in schema-enabled mode?

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+`KafkaMilvusMessageConsumer` is intentionally not a group-managed consumer. It does not subscribe, does not rely on rebalances, and does not use broker-managed group offsets for recovery decisions. Ownership and resume position are controlled by Debezium internal task state.
+
+**Key Kafka consumer configuration for correctness:**
+
+| Property | Value | Reason |
+|---|---|---|
+| `enable.auto.commit` | `false` | Debezium manages offsets via Connect storage |
+| `isolation.level` | `read_committed` | Avoid reading transactional messages mid-transaction |
+| `auto.offset.reset` | `none` | Force explicit seek; fail if offset missing |
+| `max.poll.interval.ms` | `300000` | Prevent rebalance during slow snapshot phases |
```
Naros: I don't believe this aligns with the consumer uses `assign()` strategy. When you use manual partition assignment, this configuration has no effect IIRC.

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+
+#### Wire Format Detection Decision
+
+The connector determines wire format once at startup using `MilvusWireFormatDetector`. `MarshalType` is cluster-wide and stable for a given running version; it is not expected to vary message-to-message.
+
+Startup probe algorithm:
+1. Connect to the Kafka broker.
+2. For each configured pchannel, read the first message at the next processing position:
+- earliest available message after subscription
+- message at stored seek position
```
Naros: This probe strategy assumes that there is data already in the topic. But what about situations if there are no messages or if the first message is a `TimeTick`?

DDD-42.md
Naros: I think that's reasonable, as long as we prefer the Milvus API where possible.

For places where we use etcd directly, we should evaluate whether that makes sense and whether it's an area for the Milvus API to add, if it's still something we'd need for Milvus 2.6+.

Naros (commented):
I've given this another read-through, some more observations.

DDD-42.md
kmos: agree. In this section we need to see the mapping with the Debezium messages

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+#### InsertRequest Columnar Layout (Mandatory Pivot)
+
+Milvus sends values column-wise in `FieldData`. Row `i` is reconstructed by taking index `i` from each field array.
+
+Contract:
+- Reject row if required primary key field is missing.
+- Reject batch if field column lengths do not match `num_rows`.
+- Emit one Debezium record per reconstructed row.
+
+- [msg.proto — full message definitions](https://github.com/milvus-io/milvus-proto/blob/master/proto/msg.proto)
```
kmos, line 249: about `proto` file, what is the approach in this case @Naros ? We should have a git submodule for `https://github.com/milvus-io/milvus-proto`

kmos: I just take a look to the milvus repository and the Milvus Java sdk should contains protos and apis https://github.com/milvus-io/milvus-sdk-java , @kartikangiras is it correct?

DDD-42.md
```diff
@@ -0,0 +1,759 @@
...
+**Step 2 — Build the subscription map:**
+
+
+The connector subscribes to each unique pchannel. Messages are then filtered by vchannel when processing.
+
+**Step 3 — Determine which collections to monitor:**  
+Driven by `collection.include.list` config property. If empty, monitor all collections.
+
+- [Milvus metadata API docs](https://milvus.io/api-reference/pymilvus/v2.6.x/MilvusClient/Collections/describe_collection.md)
+- [Milvus CDC overview](https://milvus.io/docs/milvus_cdc_overview.md)
```
kmos, line 348: please if the API is present in the java sdk, please add the reference

kmos (commented):
It isn't clear to me what are the libraries that you will use in the project. I think that you would have for sure the milvus java sdk and the java kafka-clients. I would underline in the document:
- external resources (libraries) to achieve this or that goal (0 risk)
- things that are inspired (go -> java) by [milvus-cdc](https://github.com/zilliztech/milvus-cdc) (low risk)
- things that you have to implement from 0 (medium risk)

kmos:
@kartikangiras please update the document with the last accordingly with the architecture and missing part before the 31th August

kartikangiras:
> @kartikangiras please update the document with the last accordingly with the architecture and missing part before the 31th August

Updated with the latest version and architecture accordingly.

kmos (approved):

PR #51, opened 2026-06-22
DDD-53: PostgreSQL LISTEN/NOTIFY Connector Mode

zavera:
## Summary

This DDD proposes a `listen-notify` mode for the Debezium PostgreSQL connector as a lightweight alternative to logical replication slots, for teams already using PostgreSQL's native `LISTEN`/`NOTIFY` mechanism.

Related issue: https://github.com/debezium/dbz/issues/2127

## Key Design Decisions

**Stateless connector instances** -- all event state lives in `debezium_outbox`. Instances carry zero local state and are fully identical and disposable.

**NOTIFY as wake-up signal only** -- `NOTIFY` triggers an immediate outbox sweep but is never the source of truth. A polling scheduler provides the correctness guarantee when `NOTIFY` is missed.

**Horizontal scaling via `FOR UPDATE SKIP LOCKED`** -- multiple instances race to claim outbox rows atomically. No coordination, no leader election, no partition assignment required.

**User-owned triggers, connector-owned function** -- the connector installs and manages `debezium_ln_notify()` only. Users attach it to their tables with whatever operations they choose. Minimises required permissions (`CONNECT` + `EXECUTE` only -- no replication privilege).

**At-least-once delivery** -- outbox rows are marked consumed only after the Debezium pipeline confirms delivery. Crash before confirm = redeliver on next sweep. Downstream deduplicates on `idempotency_key`.

## Addresses Reviewer Points from DBZ-2127

- **Trigger maintenance** -- connector manages function only; user owns trigger attachment; `CREATE OR REPLACE` makes startup idempotent
- **Connector downtime** -- outbox captures all events regardless of active listeners; polling sweeps on reconnect
- **Schema evolution** -- `row_to_json()` auto-includes new columns; breaking changes documented
- **At-least-once guarantee** -- outbox + `FOR UPDATE SKIP LOCKED` + confirm-before-consume
- **Code structure** -- `ListenNotifyConnector` extends existing `PostgresConnector` base, modelled on MySQL/MariaDB pattern


DDD-39.md
```diff
@@ -0,0 +1,505 @@
+# DDD-39: PostgreSQL LISTEN/NOTIFY Connector Mode
```
Naros: Can you open an issue in this repo, I realize it seems a bit redundant but it is to get a unique ID for this repository that you will use to name your DDD, e.g. `DDD-XX`. From there, update the PR with that information, and be sure to add an entry in the README, too. Thanks.

DDD-53.md
```diff
@@ -0,0 +1,505 @@
...
+### Trigger Maintenance
+
+The connector installs and manages `debezium_ln_notify()` only. Users manage trigger attachment.
+
+| Concern | Resolution |
+|---|---|
+| Function already exists on startup | `CREATE OR REPLACE` -- always idempotent |
+| User removes trigger externally | Events stop arriving on that channel -- no connector impact |
+| User adds new trigger at runtime | Events arrive on new channel -- connector picks up if channel is in config |
+| Multiple connector instances | Function is shared -- no naming collision; outbox ownership scoped by `connector_name` |
```
Naros: You mention here that the function is shared, allowing multiple connector instances, but elsewhere in the document you state that, on shutdown, the function is dropped. If A is stopped and drops the function, what happens to instances B and C?

So, to align with what you had here, would it be safe to say that the function is always installed, but never dropped by the connector? If so, we should update on the "on shutdown" section.

DDD-53.md
```diff
@@ -0,0 +1,505 @@
+# DDD-39: PostgreSQL LISTEN/NOTIFY Connector Mode
+
+## Motivation
+
+The existing Debezium PostgreSQL connector uses logical replication slots to capture change events. This is a robust and complete solution but carries non-trivial operational overhead: it requires a replication-enabled database user, consumes WAL resources, and mandates careful slot lifecycle management to prevent WAL accumulation.
+
+Many PostgreSQL deployments already use `LISTEN`/`NOTIFY` as a lightweight internal signalling mechanism -- particularly for event-driven architectures where row changes trigger downstream processing. For these use cases, the full logical replication machinery is disproportionate.
+
+This document proposes a `listen-notify` mode for the Debezium PostgreSQL connector using PostgreSQL's native `LISTEN`/`NOTIFY` protocol as the change capture transport. The design is built around a **stateless, horizontally scalable** connector model: all state lives in the database, every instance is identical, and no coordination between instances is required.
```
Naros: If I understand correctly, the NOTIFY payload in PostgreSQL is limited to ~8000 bytes in total. For small tables, this likely is fine, but I suspect it could easily become a hard limit on specific table configurations. This should probably be considered a significant prerequisite here because I suspect this has business transaction implications if the trigger fails due to payload size and rolls back the user transaction?

DDD-53.md
```diff
@@ -0,0 +1,505 @@
...
+            """, id);
+    }
+}
+```
+
+---
+
+#### ReconnectPolicy
+
+Uses pgjdbc-ng's `PGNotificationListener.closed()` callback, which fires when the connection drops due to Postgres restart, host reboot, or network failure. On reconnect, channel subscriptions are restored from configuration -- not from local memory -- preserving statelessness.
```
Naros: We should probably consider the impacts of needing `pgjdbc-ng` vs `pgjdbc`. 

My concern here is that there have been no releases since 2021, and contribution activity on the public repository has become nearly nonexistent in recent years. Compared to `pgjdbc`, which is actively maintained, this, to me, is a hard sell to get past.

Are there any other alternatives?

DDD-53.md
zavera: The payload is intentionally a minimal wake-up signal. Any data retrieval and transaction handling is the responsibility of the business logic in the routed function (which can carry out its own transactions), not the NOTIFY payload itself. The 8000 byte limit may be a hard requirement but it is not a concern here.

DDD-39.md
zavera: Done - opened issue [#53](https://github.com/debezium/debezium-design-documents/issues/53) to get the unique ID. The file has been renamed to DDD-53.md, the PR title updated, and a README entry added.

DDD-53.md
zavera: Valid concern - agreed, switching to pgjdbc is the right call given it is actively maintained and already used in Debezium. The tradeoff is that pgjdbc uses polling rather than real-time push for notifications, and unlike pgjdbc-ng it has no built-in reconnect callback. We would handle reconnect by catching connection failures and re-issuing LISTEN on all configured channels. This is acceptable since the outbox handles missed notifications and guarantees no data loss during any connection gap. We can update the DDD to reflect this.

DDD-53.md
zavera: Good catch - this is a documentation error in the DDD. The function should never be dropped on shutdown. It is shared database infrastructure, not owned by any single connector instance. When an instance shuts down, it closes its database connection which automatically drops its LISTEN subscriptions for that connection only. Other instances retain their own subscriptions independently and the function remains untouched in the database. We will remove the drop function step from the shutdown section.

DDD-53.md
```diff
@@ -0,0 +1,505 @@
...
+    -- row_to_json(OLD) serialises the entire row before the change.
+    -- New columns appear in the payload automatically on schema change.
+    payload := json_build_object(
+        'connector',  TG_ARGV[1],
+        'table',      TG_TABLE_NAME,
+        'op',         TG_OP,
+        'ts',         extract(epoch from clock_timestamp()),
+        'id',         gen_random_uuid(),
+        'before',     CASE WHEN TG_OP = 'INSERT' THEN NULL ELSE row_to_json(OLD) END,
+        'after',      CASE WHEN TG_OP = 'DELETE' THEN NULL ELSE row_to_json(NEW) END
```
vjuranek: How do you obtain `OLD` and `NEW`?

DDD-53.md
```diff
@@ -0,0 +1,505 @@
...
+```
+
+User-defined triggers are intentionally left intact on shutdown. The user created them; the user removes them.
+
+---
+
+#### debezium_outbox -- Single Source of Truth
+
+All event state lives here. No connector instance holds any local event state.
+
```
vjuranek, line 105: Who is in charge of maintaining this table?

vjuranek (commented):
If I'm not mistaken, the only Postgres sprecific feature is listen/notify which can be replaced by regular polling in case of other databases (and some databases may provide listren/notify equivalent). So if we accept this, I'd suggest to create it as a new general "polling outbox table" connector, with enhanced polling mechanism (using notifications) for databases which support it.

Going one step further, part of it can be generalized to "run arbitrary select from arbitrary table" which was requested by some users in the past for ad-hoc or regular snapshots.

My main concert is that we would extend the scope of Debezium project too much with it.

mfvitale:
> I'd suggest to create it as a new general "polling outbox table" connector,

I totally agree on this. Our connector was created targeting the `log-based` CDC approach. On the other hand, if I got it correctly, this implementation could be interesting for light use cases like the `outbox` pattern. 

So I agree that this should be separate connector with a `trigger/poll` based approach.

zavera:
I agree that LISTEN/NOTIFY as a PostgreSQL-specific feature is too narrow a scope. I propose reframing this as a new separate outbox polling connector that is database agnostic. The motivation is specifically deployments where logical replication slots are unavailable or undesirable ( managed databases like RDS and Cloud SQL being the primary use case.)
The proposed design:

- A Java scheduled job polls the source table on a configurable interval
- Inserts and updates are detected by comparing current table state against last known state persisted in the outbox
- The outbox serves as the state store, connector instances remain fully stateless
- No WAL, no replication slots, no special database privileges required
-
- For delete capture, we propose an optional configuration-driven extension. When enabled, the connector automatically installs a lightweight DELETE trigger on the source table that writes the OLD row directly to the outbox before it disappears. The trigger can be enabled or disabled via configuration without being dropped or recreated - a connector-owned flag in a metadata table controls whether the trigger writes to the outbox. This keeps delete capture fully optional and operationally easy to toggle.


This is not a CDC replacement -  it is a lightweight alternative specifically for teams who cannot use logical replication. For full CDC, the existing Debezium connectors remain the right tool.

DDD-53.md
zavera: Assuming the Java base layer for polling is accepted, OLD and NEW are derived by the Java polling job by comparing current table state against last known state persisted in the outbox.

DDD-53.md
zavera: Assuming the Java base layer for polling is accepted, OLD and NEW are derived by the Java polling job by comparing current table state against last known state persisted in the outbox. In this model the application owns and maintains the outbox table, with an optional configurable DELETE trigger handling the old row capture for delete events.

vjuranek: @zavera what exactly do you mean by the application? Debezium connector or the application using it?

vjuranek:
> The proposed design:
> 
>     * A Java scheduled job polls the source table on a configurable interval
> 
>     * Inserts and updates are detected by comparing current table state against last known state persisted in the outbox
> 
>     * The outbox serves as the state store, connector instances remain fully stateless
> 
>     * No WAL, no replication slots, no special database privileges required
> 
>     * For delete capture, we propose an optional configuration-driven extension. When enabled, the connector automatically installs a lightweight DELETE trigger on the source table that writes the OLD row directly to the outbox before it disappears. The trigger can be enabled or disabled via configuration without being dropped or recreated - a connector-owned flag in a metadata table controls whether the trigger writes to the outbox. This keeps delete capture fully optional and operationally easy to toggle.

Sound good to me. From the discussion above, I have a feeling that also others are fine with it, under the assumption that it's dedicated and well-isolated connector.

DDD-53.md
zavera: The Debezium connector owns and maintains the outbox table. It is created in the application's database on connector startup using the provided connection config, alongside the source tables. The application using Debezium has no direct responsibility for it.

zavera:
if I have a go based on the revised proposal mentioned by @vjuranek  then I can work towards PR

vjuranek:
@zavera I'd suggest to update the DDD accordingly first. Of course you can start working on the PR itself, but it may result into some extra cycles if we would request any further changes in the DDD design.

DDD-53.md
vjuranek: The outbox table lifecycle should be clearly described here, including possible concurrent flow when multiple connectors run in parallel, how the cleanup is done in such case, if the table is ever deleted etc.

zavera: Added a new "Outbox durability, delegated" section covering this: table restoration is delegated to whatever backup policy the operator already runs for the database as a whole. If the outbox table disappears, the scheduled job detects it on the next sweep, recreates it, and re-runs the same init routine used at t=0.

zavera:
I have updated the DDD. Essentially, this trades JVM heap for avoiding the WAL-growth risk of an unconsumed replication slot, using primitive arrays to keep the footprint small, lighter than log-based CDC on memory and slot risk.

Detection algorithm: sort-merge diff between an in-memory sorted baseline (id, checksum) and the current table state; no trigger required.
Insert: no prior baseline entry exists, so the new row becomes the baseline entry; nothing is dumped to the outbox as "old."
Update: the old row is dumped to the outbox as the change event; the new content replaces it as the new "old" baseline entry.
Delete: the old row is dumped to the outbox as the change event; its baseline entry is removed entirely, with no replacement.
Outbox dedupe: the (row_id, new_checksum) unique constraint collapses duplicate detections across replicas.
Recovery: a missing outbox is detected on the scheduled sweep, recreated, and reinitialized as at t=0.

zavera:
@vjuranek I've reworked the DDD per this thread's discussion. `LISTEN`/`NOTIFY` is gone entirely, and the doc now describes a single, standalone, database-agnostic outbox polling design. Would appreciate a look when you have a chance.

DDD-39.md
Naros: Hi @zavera the first commit still refers to the wrong DDD in the commit message, if you could take a look.

DDD-53.md
```diff
@@ -0,0 +1,283 @@
...
+---
+
+## Configuration
+
+```properties
+jdbc.url=jdbc:postgresql://host:5432/db
+jdbc.user=...
+jdbc.password=...
+table.name=orders
+table.id.column=id
```
Naros, line 226: These properties likely should be aligned to the existing PostgreSQL connector properties, e.g. `database.dbname`, `database.user`, `database.password`, `table.include.list` and so on, correct?

DDD-39.md
zavera: Fixed, the first commit now reads DDD-53 instead of DDD-39. Rewrote just that commit's subject line and rebuilt the rest of the branch on top, no content changes, force-pushed.

DDD-53.md
zavera: Deliberately not aligned. database.* naming fits the existing connector because it's permanently Postgres-specific. This design does all detection in the JVM over plain JDBC, so the database only ever sees a generic SELECT and INSERT, nothing database-specific to configure. Keeping jdbc.url/table.name generic on purpose, even though it won't match sibling connectors.

PR #54, opened 2026-06-29
debezium/dbz#52 Neo4j Debezium connector DDD

### LINKED ISSUE #52: DDD for Neo4j Connector
DDD for https://github.com/debezium/dbz/issues/2134

vsantonastaso:
closes https://github.com/debezium/debezium-design-documents/issues/52
## Summary

This DDD proposes a Debezium **source** connector for Neo4j, the first graph database connector in the Debezium ecosystem.

The connector uses Neo4j's built-in [CDC API](https://neo4j.com/docs/cdc/current/) (available since Neo4j Enterprise 5.13) to capture node and relationship changes via polling of the `db.cdc.query()` procedure, and emits them as standard Debezium envelope events to per-label and per-relationship-type Kafka topics.

## Key Design Decisions

Each link points to the relevant section in the DDD for review:

- **[Version compatibility](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R30)** : no support for Community Edition or versions prior to 5.13 (CDC API not available). Alternative approaches considered and rejected.

- **[CDC API as capture mechanism](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R44)** : uses `db.cdc.query()` with cursor-based polling. [Why polling, not push](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R58) : Neo4j CDC is poll-only by design.

- **[Topic routing](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R80)** : one topic per node label and one per relationship type. [Multi-label handling](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R98) : events emitted to all matching label topics. [Alternatives considered](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R111) .

- **[Event key resolution](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R125)** : derived from Neo4j key constraints, `null` when absent, `message.key.properties` as user override. [Alternatives considered](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R183) (line 183).

- **[Per-event schema](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R309)**: each event carries its own schema built from event properties (like MongoDB connector). [Type resolution](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R332) (line 332): types inferred from Neo4j Java Driver values per event. [No schema history topic](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R365) (line 365).

- **[Offset management](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R375)**: single `changeId` field. Each CDC change record has a unique ID and the `from` parameter is exclusive, so no `txId`/`seq` needed.

- **[Snapshot](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R393)**: Cypher-based (`MATCH` queries), 4-step process. [Alternatives considered](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R439) .

- **[Configuration](https://github.com/debezium/debezium-design-documents/pull/54/files#diff-b1a2e5c5e8e4e7c0e1f7c0e5a8b6d4c0R448)**: property names follow Debezium conventions (`node.label.include.list`, `relationship.type.include.list`, etc.).


DDD-52.md
```diff
@@ -0,0 +1,513 @@
...
+Building a Debezium connector for Neo4j should:
+
+* Bring graph databases into the Debezium ecosystem for the first time.
+* Allow users to leverage the full Debezium feature set: single message transformations, multiple sink destinations, centralized offset management, and unified monitoring.
+* Provide a consistent CDC experience for organizations running Neo4j alongside relational databases already captured by Debezium.
+
+## Goals
+
+* Provide a Debezium source connector that captures node and relationship changes from Neo4j Enterprise Edition (self-managed and Aura) databases using the built-in CDC API.
+* Support initial snapshots via Cypher queries for captured labels and relationship types.
```
mfvitale: maybe you can add a brief description/example about `Cypher`?

DDD-52.md
```diff
@@ -0,0 +1,513 @@
...
+Starting with version 5.13, Neo4j introduced a built-in [Change Data Capture (CDC) API](https://neo4j.com/docs/cdc/current/) based on enriched transaction logs.
+The CDC API exposes three Cypher procedures ([`db.cdc.earliest`, `db.cdc.current`, `db.cdc.query`](https://neo4j.com/docs/cdc/current/procedures/)) that return a cursor-based stream of committed changes with before/after state, making it a solid foundation for a CDC source connector.
+
+Building a Debezium connector for Neo4j should:
+
+* Bring graph databases into the Debezium ecosystem for the first time.
+* Allow users to leverage the full Debezium feature set: single message transformations, multiple sink destinations, centralized offset management, and unified monitoring.
+* Provide a consistent CDC experience for organizations running Neo4j alongside relational databases already captured by Debezium.
+
+## Goals
```
mfvitale, line 48: maybe could me useful to add a briefly introduction on the concept of a graph database. For example,  is there any `schema` concept?

This could help understanding the following parts.

DDD-52.md
```diff
@@ -0,0 +1,513 @@
...
+The connector will poll `db.cdc.query` on a configurable interval, passing the last processed change identifier as the `from` parameter and the configured selectors as filters.
+Each call returns a set of change records containing `id`, `txId`, `seq`, `metadata`, and `event` fields ([output schema](https://neo4j.com/docs/cdc/current/procedures/output-schema/)).
+
+#### Why polling
+
+The Neo4j CDC API is **poll-only** by design. 
+There is no way to subscribe, register a callback, or open a persistent stream that pushes events to the client. 
+The only mechanism to read changes is to call `db.cdc.query()` and get back whatever has accumulated since the last change identifier.
+
+This is different from other databases that offer push-based change streams (e.g., PostgreSQL logical replication pushes WAL records, MongoDB change streams block until new changes arrive).
```
mfvitale: Effectively also with PostgreSQL we [poll](https://github.com/debezium/debezium/blob/3f6566f3c0ea438d4c0bbb8440409a444df9765e/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java#L261) the `replication slot` , also `Oracle` do queries via `log miner`.  I don't recall the others but for sure `MySQL/MariaDB` uses a `listener` approach. 

So I think that most connector are aligned with the poll approach.

DDD-52.md
```diff
@@ -0,0 +1,513 @@
...
+| `initial` (default) | Run the snapshot first, then start streaming CDC changes. |
+| `never` | Skip the snapshot, start streaming CDC changes immediately. Useful when the consumer only cares about new changes. |
+| `initial_only` | Run the snapshot and stop. No streaming phase. Useful for one-time data exports. |
+
+The snapshot process follows these steps:
+
+**Step 1: Save the current CDC position.**
+Before reading any data, the connector calls `db.cdc.current()` to get the change identifier of the last committed transaction.
+This identifier marks the point in the CDC stream where the snapshot "started".
+The connector saves it so that after the snapshot is done, streaming can resume from this exact point without missing any changes.
```
mfvitale, line 450: Which transaction isolation neo4j uses?

vsantonastaso: Neo4j uses read-committed isolation level by default (https://neo4j.com/docs/operations-manual/current/database-internals/concurrent-data-access/)
I think that this is a minimal impact on our connector since the CDC procedures return only committed changed so we never read uncommitted data. 
The isolation level affects application transactions writing to Neo4j, not the connector reading CDC events, or at least I think so.
Any opinion about it?

DDD-52.md
vsantonastaso: Ok I was wrong. I'll adjust it, thanks 😉

DDD-52.md
vsantonastaso: I avoided a technical description of the technology used because it should be a prerequisite to reading this DDD.
However, perhaps I was wrong, so if it would be helpful, I'll add it.

DDD-52.md
vsantonastaso: Same of this https://github.com/debezium/debezium-design-documents/pull/54#discussion_r3506327727

DDD-52.md
mfvitale: The problem with `read-committed` is that you can have duplicates between snapshot phase and the streaming. 

Before starting the snapshot you get the change of last committed transaction with `db.cdc.current()` , then you start snapshotting the data via queries, but since the connection you use to read this data has `read-committed` isolation level, all changes that happens (and are committed) are then read by the snapshot phase. Since these changes happened after the identifier you get with `db.cdc.current()`, when you start streaming you will read those changes again. Ins't it?

DDD-52.md
mfvitale: Maybe just a link to Cypher could be enough.

DDD-52.md
vsantonastaso: Yes, it could certainly happen. 
Unfortunately, however, Neo4J does not provide any capability to avoid this, such as MVCC.
One possible solution would be to read the changeId after the snapshot in order to skip all changes captured during the snapshot process. However, in this case, we could lose events.
Therefore, in this initial phase, it might be better to have a duplicate event than a missing one. 
But I would like to hear your opinion because I am not sure.

DDD-52.md
```diff
@@ -0,0 +1,513 @@
...
+* Support initial snapshots via Cypher queries for captured labels and relationship types.
+* Map Neo4j graph entities (nodes and relationships) to Kafka topics following Debezium conventions.
+* Produce change events with before/after state that conform to the Debezium envelope structure.
+* Provide offset tracking that survives connector restarts without data loss.
+* Support filtering by node labels, relationship types, and properties.
+
+## Out-of-scope
+
+* Backward compatibility with Neo4j Community Edition or versions prior to 5.13: the connector requires the [CDC API](https://neo4j.com/docs/cdc/current/), which is only available in Neo4j Enterprise Edition 5.13+. Community Edition does not expose transaction log enrichment or the `db.cdc.*` procedures. Earlier Enterprise versions do not have a built-in CDC capability. Alternative approaches (APOC triggers, timestamp-based polling, direct transaction log parsing) were considered but rejected because they cannot guarantee at-least-once delivery, do not provide before/after state, and would require a fundamentally different connector architecture.
```
mfvitale, line 59: This could be a limitation but on the other hand using the `cdc` feature could be an easy way to start familiarizing with this kind of databases, and, if then there is more demand for the community, we could investigate if it is possible to directly access the log.

DDD-52.md
vsantonastaso: Added a section with key concepts of Neo4J. Thanks 😉

DDD-52.md
vsantonastaso: Done. Thanks 😉

DDD-52.md
vsantonastaso: I agree completely. 
For the first iteration, we could accept this limitation and gain a better understanding of whether we need to add this complexity based on community feedback.

vsantonastaso:
On hold, as we are evaluating an alternative approach of implementing only the SMTs rather than the entire connector.

PR #56, opened 2026-07-22
debezium/dbz#55 Add DDD for Relational-to-Graph Debezium Source SMT for Neo4j

### LINKED ISSUE #55: DDD for Neo4j SMT relational-to-graph

vsantonastaso:
Closes: DDD-55 (#55)
 
## Summary

This DDD proposes a Kafka Connect SMT that transforms Debezium change events into Neo4j's CUD format
This is the result of the team discussion where this SMT solution was chosen over building a full connector. 
The DDD focuses on the **Relational -> Graph** direction (V1). 
The reverse direction (Graph -> Relational) is deferred to a separate DDD.

  ## Key Design Decisions

  - **CUD as target format**: CUD was chosen over Cypher/Pattern/CDC because it maps directly to Debezium operations (c/r/u/d -> merge/delete) without requiring user-written Cypher queries

  - **Explicit user configuration**: V1 requires the user to declare the graph structure (labels, keys, FK relationships) explicitly.

  - **Operation mapping**: all create/update/snapshot operations use `merge` (not `create`) for idempotency with at-least-once delivery

  - **Join table support**: `node.mode=relationship_only` suppresses node events for join tables, producing only relationship CUD events

  - **Two-instance approach for multi-event output**: since the Kafka Connect `Transformation<R>` interface returns a single record, tables that need both node and relationship events require two sink connector instances. Edge cases documented


DDD-55.md
```diff
@@ -0,0 +1,665 @@
...
+#### Multiple CUD Events per Debezium Event
+
+The standard Kafka Connect `Transformation<R>` interface returns a single record from `apply(R record)`. 
+The SMT transforms a record in-place before the sink connector writes it to Neo4j. 
+This means one Debezium event can only produce one CUD event.
+
+For entity tables that also have FK columns configured as relationships, the SMT needs to produce both a node event and one or more relationship events from a single Debezium event. 
+Since the `apply()` method can only return one record, this requires a workaround.
+
+**V1 approach: two sink connector instances per topic**
```
Naros: Should we also consider the SMT creating a single Kafka record that contains the node and relationship (packed as a JSON array or similar), and the SMT is applied on the source side of the pipeline, writing to an intermediate topic/target, e.g. `[{node CUD}, {relationship CUD}, {relationship CUD}]`.

Then a ksqlDB query could fan this out using explode, e.g.:
```
CREATE STREAM neo4j_cud_events AS
  SELECT EXPLODE(cud_events) as cud_event
  FROM neo4j_intermediate
  EMIT CHANGES;
```
I think this may address a number of edge cases raised below?

vsantonastaso: mmm yes, it may address the edge cases mentioned, but I have doubts about the infrastructure dependency that we would add to the pipeline. 
The solution requires either ksqlDB or a Kafka Streams app as a mandatory component. 
Since ksqlDB is a Confluent product, users of vanilla Apache Kafka would need to write a custom Kafka Streams app instead.
I don't know if we already have something similar in place for other connectors.
Otherwise we could accept the limitation for the V1 and provide a stateful solution for V2 that will also address the automatic mapping inference without the need for explicit and verbose configuration.

vsantonastaso: Actually, thinking about it more, I like this approach. We could implement it as the default behavior of the SMT.
Basically the SMT runs on the Debezium **source** connector side and always packs all CUD events (node + relationships) into a JSON array in a single Kafka record. For a table  with no relationships, the array could contain a single element.

But in this case is the user's responsibility using whatever stream processing tool they have available. We can document _ksqlDB EXPLODE_ as the simplest example, but Kafka Streams, Flink, or any other tool that reads from Kafka would work.
This eliminates the two-instance approach entirely, along with all the race condition edge cases. 
The SMT stays stateless and produces a single record per Debezium event. The ordering is guaranteed because node and relationship events are packed in the same record.
I'll update the DDD to reflect this. WDYT? @Naros @jpechane

jpechane: @vsantonastaso I think it is fine to have it supported in SMT but I am not sold on this beign the default behaviour. With this as default would the SMT be directly consumable by neo4j sink?

jpechane:
@vsantonastaso Excellent job! I really like how you thought through different details specific for graph mapping.

DDD-55.md
vsantonastaso: @jpechane no, with array mode as default, the output would not be directly consumable by the Neo4j sink connector. 
The user would always need a way to split the array (ksqlDB, Kafka Streams, etc.) even for tables with no relationships.
To address this, I see two possible approaches:

1. a single SMT class deployed on the Debezium source connector, with a configuration property to control the output. A default could produce one CUD event per Debezium event. For tables with FK relationships, only the node is emitted. Relationships require a separate configuration (as documented in the edge cases). The Neo4j sink connector can consume the output directly.
Another configuration could split all CUD events (node + relationships) into a JSON array and this will requires a  step before the Neo4j sink connector, but eliminates ordering/race condition edge cases.

2. two separate SMTs, a source-side SMT that produces the packed JSON array and a sink-side SMT that produces a single CUD event per record (directly consumable by the Neo4j sink connector)
WDYT? maybe we should keep it simple and just go with the array mode, accepting the fan-out step as a requirement for the V1, but I am not sure

vsantonastaso:
> @vsantonastaso Excellent job! I really like how you thought through different details specific for graph mapping.

Thanks, I really appreciate it

DDD-55.md
mfvitale: I vote for option 1.

DDD-55.md
```diff
@@ -0,0 +1,665 @@
...
+# --- Customers (simple node, no relationships) ---
+transforms.neo4j_customers.node.labels=Customer
+transforms.neo4j_customers.node.id.properties=id
+
+# --- Orders (node + relationship to Customer) ---
+transforms.neo4j_orders.node.labels=Order
+transforms.neo4j_orders.node.id.properties=id
+transforms.neo4j_orders.relationship.customer_id.type=PLACED_BY
+transforms.neo4j_orders.relationship.customer_id.target.label=Customer
+transforms.neo4j_orders.relationship.customer_id.target.id=id
```
mfvitale, line 676: Based on what described in `Multiple CUD Events per Debezium Event` how this is going to work? Should there be a configuration for the node and one for the relation?

DDD-55.md
```diff
@@ -0,0 +1,665 @@
...
+
+2. **Match vs merge tradeoff**: setting `from.op: "match"` avoids the stale merge problem but causes relationship creation to fail when the relationship instance runs ahead of the node instance (the node does not exist yet). There is no setting that is safe in both directions.
+
+3. **Detach removes relationships from the other instance**: the node instance processes a DELETE with `detach=true` and removes all relationships, including ones that the relationship instance just created moments before from a different event.
+
+4. **Consumer rebalancing**: Kafka can rebalance consumer groups independently for the two instances, causing their offsets to drift further apart and widening the window for the race conditions above.
+
+These edge cases are related to running two independent consumers over the same topic. 
+For schemas where nodes and relationships must be strictly consistent (e.g., financial data), a stateful sidecar (see [Future Work](#future-work)) that coordinates node and relationship creation within a single consumer is the
```
mfvitale: Is the `Future Work` section missing?

mfvitale:
@vsantonastaso Overall it LGTM. I left some doubts.

DDD-55.md
vsantonastaso: Yes, it's a typo. I originally called that section 'Future Work', but then I renamed it 'Out of Scope' to avoid making promises.

DDD-55.md
jpechane: Option 1 is IMHO better

vsantonastaso: Thanks @Naros @mfvitale @jpechane, Let's do the option 1.
Just to summarize, the SMT will be a single class deployed on the Debezium source connector with a configuration property (e.g. output.mode) that controls the output format:
  - output.mode=single (default): produces one CUD event per Debezium event. For tables with FK relationships, only the node event is emitted and FK columns are excluded from properties. Directly consumable by the Neo4j sink connector.
  - output.mode=array: packs all CUD events (node + relationships) into a JSON array in a single record snf requires a split step with ksqlDB or Kafka Streams (in charge of the users) before the Neo4j sink connector

I'll update the DDD.

DDD-55.md
vsantonastaso: @mfvitale good catch, the configuration is confusing as written but since I'm going to change the approach and update the DDD I will rewrite this part 👍

vsantonastaso:
@jpechane @Naros @mfvitale I updated the DDD based on your feedback. Here's what changed.
 The SMT now runs on the source connector side instead of the sink connector.

The SMT now supports two output modes. The default is **output.mode=array,** which packs all CUD events (node + relationships) into a JSON array in a single Kafka record. This guarantees that nodes are always created before their relationships but it requires a split step before the Neo4j sink
  connector can consume the output. I added a dedicated section 

The alternative is **output.mode=single**, which produces one CUD event per record and is directly consumable by the Neo4j sink connector. The limitation here is that it can only emit either a node or a relationship, not both from the same event.

Let me know WDYT

DDD-55.md
```diff
@@ -0,0 +1,827 @@
...
+| `FLOAT64` | `Double` | |
+| `BOOLEAN` | `Boolean` | |
+| `STRING` | `String` | |
+| `BYTES` | `ByteArray` | |
+| `io.debezium.time.Date` | `String` (ISO date) | Converted to ISO-8601 date string |
+| `io.debezium.time.Timestamp` | `String` (ISO datetime) | Converted to ISO-8601 datetime string |
+| `io.debezium.time.MicroTimestamp` | `String` (ISO datetime) | Converted to ISO-8601 datetime string |
+| `io.debezium.time.ZonedTimestamp` | `String` (ISO datetime) | Preserves timezone information |
+| `io.debezium.data.Json` | `String` | Stored as JSON string |
+| `org.apache.kafka.connect.data.Decimal` | `Double` | Precision loss possible for very large decimals |
```
Naros: Perhaps there should be a note in the docs that highlight source connectors should either opt to set `decimal.handling.mode` to either `double` or `string` based on how they want the data to be serialized through the SMT? That way for situations like `precise` using `VariableScaleDecimal`, the SMT wouldn't need to support it?

What do others think?

vsantonastaso: I didn't know about this configuration before so thanks @Naros.
I agreed, I'll add a note recommending that users set _decimal.handling.mode_ to either double or string on the source connector depending on their precision needs (double if they need numeric operations in Neo4j or 
string if they need exact precision and don't need numeric operations)
My doubts is when the user keeps the default precise mode, I think the SMT should fail with a clear error message rather than silently converting the value with and eventual precision loss.

  WDYT?

Naros: So IIRC this depends on `errors.tolerance` in Kafka Connect, and when it's `all`, any errors that are triggered from within the SMT layer the message is skipped and not delivered, but it won't stop the producer; it continues sending events -- so it's effectively a data loss scenario in this case.

I am not sure there's much we can do here either because `transforms.*` properties are not supplied to the producer, so we can't check if the SMT is paired with the wrong `decimal.handling.mode`, and the `decimal.handling.mode` property isn't given to the transform's `configure` method :disappointed: 

So I guess that leaves making sure this is clearly documented.

Naros: Actually, `Connector#validate` gets all the data, so we could validate it there. I think that gets passed to all the validation handlers on the `Field`, so adding this on the `decimal.handling.mode` field might work. It's worth a test.

vsantonastaso: I'm not sure about adding that validation to `RelationalBaseSourceConnector.validate`. If I do, I'll end up with Neo4J specific logic for all relational connectors. I would avoid it. The problem here is that we have a validation related to an SMT, but based on properties that are not SMT related.

vsantonastaso: Maybe for this first version, we could keep the current `ConnectException` on `VariableScaleDecimal` since it  already works for errors.tolerance=none (the default). 
For errors.tolerance=all, we could document the silent data loss risk clearly. WDYT?
Another option could be add a generic hook in `RelationalBaseSourceConnector` that any SMT can use to validate config props, but that's an overkill change for V1

mfvitale: Just a stupid question. Why we cannot support `VariableScaleDecimal`?

vsantonastaso: Unfortunately, we cannot support BigDecimal/VariableScaleDecimal because Neo4j properties support Long, Double, String, Boolean, and arrays of those. 
There is no special type of decimal number that handle these precision types. 
To avoid losing precision without warning, I think it's best to suggest using a `different decimal.handling.mode `configuration.

mfvitale: Could we just fallback to `string` in case of high precision and document it?

vsantonastaso: I'm not sure. You can set the `decimal.handling.mode` configuration to treat decimals as strings.
If a user sets that property differently, we should avoid using a string silently, in my opinion

mfvitale: That's correct. But if I'm not wrong if an user set `decimal.handling.mode=string` you will not get a `VariableScaleDecimal`  right?

vsantonastaso: After a quick chat with @mfvitale , We thought it would be best to keep things simple. The `errors.tolerance` setting is usually set to `none`, so it might be annoying for the user if the message is rejected when `decimal.handling.mode` is set to `precise`. 
If, instead, we fallback to string in case `decimal.handling.mode`= `precise`, the user  won't miss any messages or break connector. 
We could document it to explain this behaviour. 
For sure both of these approaches  are valid, but the fallback logic is probably easier to use and less invasive than the message rejection or the connector fatal exception.
@Naros WDYT?

PR #58, opened 2026-07-28
DDD-57 DDD for Debezium Platform Alerting

mfvitale:


DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+## Proposed Solution: Platform-Native Alerting Engine
+
+### Overview
+
+The alerting system is embedded entirely within the Conductor (Quarkus backend). It extends the existing monitoring architecture by adding:
+
+1. **Alert rules**: user-defined threshold conditions referencing existing monitoring panels
+2. **Evaluation engine**: a periodic loop that queries Prometheus and compares results against thresholds
+3. **State machine**: tracks alert lifecycle per rule-pipeline pair with persistence in PostgreSQL
+4. **Notification subsystem**: dispatches alerts through email and webhook channels
```
indraraj, line 67: Is in-app notification not covered in the current scope?

mfvitale: in-app is the default.

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+
+Each fire/resolve cycle produces one **incident** row (`alert_event`). The row is created when the alert fires (with `resolved_at = NULL`) and updated with a resolution timestamp when the alert clears. If the same rule fires and resolves multiple times for the same pipeline, each cycle produces a separate incident row, preserving full history.
+
+### Severity Levels
+
+Each rule has a severity: **Critical**, **Warning**, or **Info**. Severity determines visual treatment and badge behavior:
+
+| Severity | Color | Badge |
+|----------|-------|-------|
+| **Critical** | Red (`--pf-v5-global--danger-color--100`) | Counted in alert badge |
```
indraraj: What is badge here referring to?

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+| **Warning** | Amber (`--pf-v5-global--warning-color--100`) | Counted in alert badge |
+| **Info** | Blue (`--pf-v5-global--info-color--100`) | Not counted in badge |
```
indraraj: Should not only be bound to colour but also differ in visual elements. We could mention PF severity icons `rh-ui-severity-critical-fill`, `rh-ui-severity-moderate-fill` and `rh-ui-severity-none-fill` as they utilize color and visual weight to reflect the sense of severity
 https://www.patternfly.org/patterns/status-and-severity

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+|--------|------|-------------|
+| `GET` | `/api/alerts/rules` | List all alert rules |
+| `GET` | `/api/alerts/rules/{id}` | Get a rule by ID |
+| `POST` | `/api/alerts/rules` | Create a new rule |
+| `PUT` | `/api/alerts/rules/{id}` | Update a rule |
+| `DELETE` | `/api/alerts/rules/{id}` | Delete a rule (204) |
+| `PUT` | `/api/alerts/rules/{id}/enable` | Enable a rule |
+| `PUT` | `/api/alerts/rules/{id}/disable` | Disable a rule |
+
+**Create/Update request body:**
```
indraraj, line 269: Not sure if this is a valid use case, but could there be scenarios where we may want to define a rule for a specific pipeline only?

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+      "severity": "CRITICAL",
+      "value": 15.2,
+      "threshold": 10.0,
+      "since": "2026-07-21T18:00:12Z"
+    }
+  ]
+}
+```
+
+The UI uses `GET /api/alerts/status` to drive the global alert badge. Poll this endpoint periodically (e.g., every 30s) to keep the badge count current.
```
indraraj, line 471: Yes, but SSE will serve better here

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│  Debezium Platform            [Alerts 🔔 3]    [User ▼]    │
+└─────────────────────────────────────────────────────────────┘
+```
+
+- Shows the count of currently **firing** Critical and Warning alerts (Info excluded)
+- Badge color reflects the highest active severity: red if any Critical, amber if only Warning
+- Clicking the badge navigates to the alert history page filtered to `status=FIRING`
```
indraraj, line 504: I am thinking: should we use the global notification badge or have this badge on the `Alerts` in the side navigation menu?

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+│                                                       │
+│  Name:        [ high-source-lag                    ]  │
+│  Description: [ Alert when source lag is too high  ]  │
+│                                                       │
+│  ── Condition ──────────────────────────────────────  │
+│                                                       │
+│  When  [ Source Lag          ▼]   (panel picker)      │
+│        [ last                ▼]   (reduce function)   │
+│  is    [ greater than        ▼]   (operator)          │
+│        [ 10.0               ]     (threshold)         │
```
indraraj, line 549: Will it be a free-text field?

DDD-57.md
mfvitale: It's the notification icon on top.

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+The alerting system is embedded entirely within the Conductor (Quarkus backend). It extends the existing monitoring architecture by adding:
+
+1. **Alert rules**: user-defined threshold conditions referencing existing monitoring panels
+2. **Evaluation engine**: a periodic loop that queries Prometheus and compares results against thresholds
+3. **State machine**: tracks alert lifecycle per rule-pipeline pair with persistence in PostgreSQL
+4. **Notification subsystem**: dispatches alerts through email and webhook channels
+5. **Incident history**: records every fire/resolve cycle for audit and investigation
+
+No new infrastructure is introduced. The system reuses Prometheus (already required for monitoring) and PostgreSQL (already the platform's persistence layer).
+
```
pxcamus, line 71: I was wondering as per https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness. What happens if a pod dies after reporting for a while? I.e. what if metrics are missing altogether? How would the evaluation loop behaves?

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+```json
+{
+  "name": "slack-webhook",
+  "type": "WEBHOOK",
+  "config": {
+    "url": "https://hooks.slack.com/services/T.../B.../xxx",
+    "method": "POST",
+    "headers": {
+      "Authorization": "Bearer token"
```
pxcamus, line 359: Those headers values are secret. Should they be write-only on input and masked in responses? And what about data at rest?

DDD-57.md
```diff
@@ -0,0 +1,688 @@
...
+  "enabled": true
+}
+```
+
+```json
+{
+  "name": "slack-webhook",
+  "type": "WEBHOOK",
+  "config": {
+    "url": "https://hooks.slack.com/services/T.../B.../xxx",
```
pxcamus, line 356: In term of best practices, I would recommend at least documenting basic security measures regarding the URL:

- Block private or loopback addresses. Since the HTTP verb runs inside a privileged network, this ensure no port scanning on what could be sensitive resources.
- DNS resolution check for every call, to ensure it resolves to a public IP.
- We could maintain a whitelist of SaaS receivers, and for deployments where webhooks are internal, require an operator config through the chart -- otherwise it is easily bypassed.

DDD-57.md
mfvitale: The `threshold`, yes.

DDD-57.md
mfvitale: I really do know. 

Having it globally as the advantages that 

* new pipelines are automatically covered
* simple UX and reduced possibility of missing a pipeline

with the disadvantages that there is no possibility to differentiated thresholds per workload type, disable temporary a rule for a specific pipeline. On the other hand users need to set fine-grain rules with a possibility rule explosion. 

Since  per-pipeline scoping is a refinement that becomes valuable as deployments grow and the cost to add it later is near-zero,  I'd defer it but document clearly that it's planned and architecturally supported.

WDYT?

@indraraj @jpechane @Naros

DDD-57.md
mfvitale: I'll try to use the global one, maybe with an inside section? Effectively it's underutilized. We can then always move it later if we feel it's better. WDYT?

DDD-57.md
mfvitale: This is listed as a non goal that can be implemented in the future if we found that real-time become a need. 
If users wants to be notified in real-time they can use the webhook for email ( I think that most production users will use then instead of in-app).  WDYT?

A part from that, if the SSE simplify the FE we can think to add it now.

DDD-57.md
mfvitale: Well this is a good point. 

The idea is to keep the status as if. So if a pipeline is in status `FIRING` if no data returned it remains in failing, same for `PENDING`.

 When the pipeline comes back:
  - If the metric is above threshold and state is PENDING, the forDuration clock continues from pendingSince
  - If the metric is above threshold and state is FIRING, it stays FIRING (no duplicate notification)
  - If the metric is below threshold, both PENDING and FIRING resolve normally

This helps to preserve the state before the die and eventually help with a root cause analysis. 

WDYT? 

In future we could thing to add a setting so user can decide the behavior when no data: reset status, keep, alert.

DDD-57.md
mfvitale: > Should they be write-only on input and masked in responses?

Yes. 

> And what about data at rest?

Could you elaborate more?

pxcamus: If the `Authorization` header lands as clear-text in PosgtreSQL, it becomes at risk for anything with read access: a raw dump, replica/backup/snapshot, etc.. 

To prevent scope creep, I would recommend we have the conductor call the Kubernetes API to create a `Secret`, and only store a ref to it.

A better approach is to externalize this to a Vault, but that will take some time. I am hoping to have a working demo with a month or so FYI.

mfvitale: Ok got the point and you are totally right. But this is also true also for connections, etc. All this problem should be then resolved once we have the `Vault` in place. So it is out of scope.

pxcamus: Sounds good, I started the prep work for that.

DDD-57.md
mfvitale: All good point. 

For an initial phase I think we can aim for:

- SSRF validation in the webhook notifier (block private/loopback/link-local IPs at call time)
- One Helm value: `alerting.webhook.allowPrivateNetworks`: false

WDYT?

pxcamus: LGTM

DDD-57.md
indraraj: Yes, I agree with underutilised part of the global notification. Sure, we can analyse later once we have other things on the global notification

DDD-57.md
mfvitale: As discussed in the weekly forum we will go for `global` rules as start and then adapt if required by community.

PR #60, opened 2026-07-28
DDD-59: Generalize per-table/per-chunk snapshot retry (dbz#2297)

### LINKED ISSUE #59: DDD: Generalize per-table/per-chunk snapshot retry (dbz#2297)
## Summary

Proposing a Debezium Design Document to generalize per-table/per-chunk retry during the initial snapshot, addressing debezium/dbz#2297.

Today a single transient `SQLException` on one table or chunk fails the whole initial snapshot. The only recovery, `errors.max.retries`, restarts it from the beginning — re-reading already-completed tables (topic duplicates on large datasets) and unwinding the offset / consistent-read setup.

The Oracle connector already solves this connector-locally via `snapshot.database.errors.max.retries` (narrow ORA-01466 retriability). This DDD proposes lifting the mechanism into the common `RelationalSnapshotChangeEventSource` for all relational connectors, and resolving what a connector-agnostic version must handle that the Oracle-only one could sidestep:

- retriability as an overridable predicate (not "any SQLException"), reconciled with Oracle's existing property;
- connection recovery after an aborted transaction (rollback + reconnect, reapplying connector-specific per-connection state);
- honest duplicate / blast-radius semantics (snapshot rows dispatch per-row mid-scan, so a retry re-emits — reduced blast radius, not idempotent; plus the `SnapshotRecord.FIRST` marker case);
- per-connector consistency (Oracle `AS OF SCN`, MySQL global lock, Postgres exported-snapshot pinning).

Opening this to reserve a DDD number; the full `DDD-<N>.md` PR will follow.

Refs: debezium/dbz#2297

TimurRakhmatullin86:
Design document for debezium/dbz#2297, tracked as DDD-59 (debezium/debezium-design-documents#59).

Generalizes the per-table/per-chunk snapshot retry that the Oracle connector already implements locally (`snapshot.database.errors.max.retries`) into the common `RelationalSnapshotChangeEventSource`, and works through what a connector-agnostic version has to handle that the Oracle-only one could sidestep — overridable retriability, connection recovery after an aborted transaction, honest duplicate/blast-radius semantics (rows dispatch mid-scan), and per-connector consistency (Oracle `AS OF SCN`, MySQL global lock, Postgres exported-snapshot pinning).

Open questions are listed at the end of the document; feedback on the config-reconciliation with Oracle's existing property and on the Postgres multi-connection consistency scope especially welcome.

Closes debezium/debezium-design-documents#59

### DIFF
--- a/DDD-59.md
+++ b/DDD-59.md
@@ -1,49 +1,44 @@
 # DDD-59: Generalizing per-table / per-chunk snapshot retry
 
 Tracking issue: [debezium/dbz#2297](https://github.com/debezium/dbz/issues/2297).
-Incorporates review feedback from @Naros on PR #60 (resume-from-key, whole-snapshot budget,
-constant backoff default, deprecated-alias for the Oracle property, Postgres pinning split out).
 
 ## Motivation
 
-During the initial snapshot Debezium reads every captured table through a fixed-size thread pool,
-splitting large tables into chunks when configured
-(`RelationalSnapshotChangeEventSource#createDataEvents`). When one table or chunk read throws a
-`SQLException` — a connection reset partway through a large table, for instance — the callable wraps
-it in a `ConnectException`, it escapes the `ExecutorCompletionService`, and `doExecute` fails the
-whole snapshot.
+During the initial snapshot Debezium reads every captured table through a fixed-size thread pool, splitting large tables into chunks when configured (`RelationalSnapshotChangeEventSource#createDataEvents`).
+When one table or chunk read throws a `SQLException` — a connection reset partway through a large table, for instance — the callable wraps it in a `ConnectException`, it escapes the `ExecutorCompletionService`, and `doExecute` fails the whole snapshot.
 
-The only recovery today is the connector-level `errors.max.retries`, which restarts the snapshot
-from the beginning. On a large dataset that is the wrong unit of recovery: it re-reads the tables
-and chunks that already succeeded (topic duplicates), and unwinds all the way back to where the
-offset and the consistent-read position were established. A chunk that fails for a transient reason
-should be recoverable by retrying *that chunk*.
+The only recovery today is the connector-level `errors.max.retries`, which restarts the snapshot from the beginning.
+That makes full snapshot re-consistency the only recovery mode on offer.
+A user who would happily accept eventual consistency — re-read just the failed unit and let consumers dedupe on key — has no way to opt into a cheaper recovery.
+That flexibility is the real motivation here: the whole-snapshot restart is expensive precisely because it insists on more consistency than many deployments actually need.
+The cost is concrete on a large dataset.
+The restart re-reads the tables and chunks that already succeeded (topic duplicates), and it unwinds all the way back to where the offset and the consistent-read position were first established.
+A chunk that fails for a transient reason should instead be recoverable by retrying *that chunk*.
 
 ## Prior art
 
-The Oracle connector already does this connector-locally: `OracleSnapshotChangeEventSource` overrides
-`createDataEventsForTableCallable` with a bounded retry loop gated by
-`snapshot.database.errors.max.retries` (`int`, default `0`), with a narrow retriability check
-(`isTableSnapshotErrorRetriable` — only ORA-01466), backing off with `Metronome`, firing
-`notifyCompletedTableWithError` per attempt, hand-rolled rather than via `createPooledResourceCallable`.
+The Oracle connector already does this connector-locally.
+`OracleSnapshotChangeEventSource` overrides `createDataEventsForTableCallable` with a bounded retry loop gated by `snapshot.database.errors.max.retries` (`int`, default `0`), with a narrow retriability check (`isTableSnapshotErrorRetriable` — only ORA-01466), backing off with `Metronome`, firing `notifyCompletedTableWithError` per attempt, hand-rolled rather than via `createPooledResourceCallable`.
+
+Oracle can afford to retry a single table read this cheaply because of *how* it pins the snapshot.
+The snapshot phase uses Flashback queries (`SELECT … AS OF SCN`), so every thread reads each table at the same SCN with no shared lock, transaction, or other server-side resource that has to survive across a retry.
+A retried read at the same SCN simply returns identical data, so re-running one table is safe and self-consistent without any extra bookkeeping.
+Most other connectors cannot sidestep the problem this way — see [Consistency per connector](#consistency-per-connector) — which is exactly what a connector-agnostic version has to account for.
 
 So the question isn't whether snapshot reads should be retriable — one connector already decided yes.
-It's whether that belongs in the common snapshot layer for every relational connector, and what a
-connector-agnostic version has to get right that the Oracle-only one could sidestep.
+It's whether that belongs in the common snapshot layer for every relational connector, and what a connector-agnostic version has to get right that the Oracle-only one could sidestep.
 
 ## Goals
 
-- A common, opt-in per-table/per-chunk retry in `RelationalSnapshotChangeEventSource`, with a
-  connector-overridable retriability predicate.
+- A common, opt-in per-table/per-chunk retry in `RelationalSnapshotChangeEventSource`, with a connector-overridable retriability predicate.
 - Default to current behavior; `0` retries changes nothing.
-- Reconcile with Oracle's existing `snapshot.database.errors.max.retries` via a deprecated alias
-  rather than a second overlapping knob.
-- Retry by **resuming from the last-emitted key** rather than re-emitting a chunk, so retry does not
-  produce duplicate rows or duplicate `FIRST`/`LAST` markers — using the mechanism the incremental
-  snapshot already relies on.
+- Reconcile with Oracle's existing `snapshot.database.errors.max.retries` via a deprecated alias rather than a second overlapping knob (see [Backward compatibility](#backward-compatibility)).
+- Retry by **resuming from the last-emitted key** rather than re-emitting a chunk, so retry does not produce duplicate rows or duplicate `FIRST`/`LAST` markers — using the mechanism the incremental snapshot already relies on.
 
-Non-goals: changing `errors.max.retries` (whole-snapshot retry still sits above this); resuming
-across a connector restart (v1 keeps the high-water key in memory for same-run retry — see caveats).
+Non-goals:
+
+- Changing `errors.max.retries` — the whole-snapshot retry still sits above this.
+- Resuming a snapshot across a connector restart — explicitly out of scope; ad-hoc blocking and incremental snapshots already cover that (see caveats).
 
 ## Proposed changes
 
@@ -51,25 +46,18 @@ across a connector restart (v1 keeps the high-water key in memory for same-run r
 
 | Property | Type | Default | Meaning |
 |---|---|---|---|
-| `snapshot.retries.max` | int | `0` | Times the snapshot may retry a retriable table/chunk failure, as one **whole-snapshot** budget. `0` = current behavior. |
+| `snapshot.retry.max` | int | `0` | Times the snapshot may retry a retriable table/chunk failure, as one **whole-snapshot** budget. `0` = current behavior. |
 | `snapshot.retry.delay.ms` | long | `10000` | Delay between attempts. |
 
-Per @Naros, this is a single whole-snapshot budget rather than per-unit — simpler contract, and
-per-unit granularity can be added later if needed (easier than walking it back). Following house
-style (`SNAPSHOT_LOCK_TIMEOUT_MS` / `snapshotLockTimeout()` returning `Duration`), the delay getter
-returns a `Duration` and the max field is `Field::isNonNegativeInteger`-validated.
+Both properties share the single `snapshot.retry` stem (`snapshot.retry.max` / `snapshot.retry.delay.ms`) so they read as one group rather than as two unrelated knobs.
 
-**Reconciling with Oracle.** The common property becomes canonical; Oracle's
-`snapshot.database.errors.max.retries` is wired as a deprecated alias via
-`Field.withDeprecatedAliases(...)` (the same mechanism Oracle already uses for
-`DEPRECATED_XSTREAM_SERVER_NAME`), so existing Oracle configs keep working and get the deprecation
-warning, and the alias can be dropped after a release or two. Oracle keeps only its retriability
-override (ORA-01466), not the whole loop.
+This is a single whole-snapshot budget rather than per-unit: a simpler contract, and per-unit granularity can be added later if it's needed, which is easier than walking granularity back once it's been given.
+Following house style (`SNAPSHOT_LOCK_TIMEOUT_MS` / `snapshotLockTimeout()` returning `Duration`), the delay getter returns a `Duration` and the max field is `Field::isNonNegativeInteger`-validated.
 
 ### Retriability, not "any SQLException"
 
-Retrying every `SQLException` would burn the budget on permanent failures (bad credentials, a broken
-`snapshot.select.statement.overrides`, a revoked grant). The loop consults an overridable predicate:
+Retrying every `SQLException` would burn the budget on permanent failures (bad credentials, a broken `snapshot.select.statement.overrides`, a revoked grant).
+The loop consults an overridable predicate:
 
 ```java
 protected boolean isSnapshotErrorRetriable(SQLException e) {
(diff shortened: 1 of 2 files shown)


DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+row of a chunk fires `signalFirstRecordEmitted()` mid-scan; the underlying `CountDownLatch` signals
+are idempotent, but the *emitted event* is not. If the chunk carrying the snapshot's first row fails
+after emitting it and is retried, the retry emits a second `FIRST`-tagged record. The loop therefore
+needs to gate marker emission on retried units (only emit `FIRST`/`LAST` on the first attempt), or we
+accept and document duplicate markers — flagged in Open questions.
+
+### Consistency is per-connector, not universal
+
+Whether a retry in a new transaction drifts in time depends entirely on how the connector pins the
+read, and it varies:
```
Naros: It would be good to document other connectors here as you are able, MariaDB, SQL Server, Db2, etc.

DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+7. Tests: a chunk failing N−1 times then succeeding completes; N+1 fails with the attempt count;
+   `0` reproduces current behavior; a non-retriable error fails immediately without consuming the
+   budget; `InterruptedException` aborts regardless; retry does not emit a second `FIRST` marker.
+8. Docs: the properties, the duplicate/blast-radius note, and the per-connector consistency and
+   availability contracts.
+
+## Open questions
+
+- Oracle reconciliation: deprecated alias mapping `snapshot.database.errors.max.retries` onto the new
+  property, or keep both for a release with the Oracle one delegating?
```
Naros: Add the `deprecatedAlias` to the new mappings so that the old Oracle name maps cleanly to the new name without extra code, and users automatically get the deprecated warning. We can then decide in 1 or 2 releases when it's reasonable to remove the alias deprecation.

DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+- Marker duplication: gate `FIRST`/`LAST` on first attempt, or document that retried units may repeat
+  them?
```
Naros: This raises an interesting point, and I'm not sure how complex this may make things, so bear with me.

The way the while loop is currently structured, it centers on the data dispatch aspect. For snapshots, we could consider sorting the rows in ascending order, even in single-threaded mode, caching the previous PK for each row read, and, on retry, fetching the data again while filtering out any rows we had already sent. This would solve the `FIRST`/`LAST` concern, but we'd need to think through the provenance of source info timestamps.

For chunked and legacy single-table-per-thread snapshots, we could do precisely the same thing I believe, minimize data duplicates and avoid any concerns with markers.

WDYT?

DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+- Retry budget per-unit (proposed) or shared across the whole snapshot?
```
Naros: Personally, I'd favor the whole snapshot, and if we need more granularity, we can always work toward that. It's always more difficult to walk back granularity after it's given.

DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+- Constant vs. exponential backoff — `DelayStrategy` supports both.
```
Naros: I can see the advantages of both.

My concern is the impact of exponential on connectors where snapshots have a limited availability window before you are forced to start over, e.g., Oracle. Here, you generally have about 15 minutes OOTB if the DBA hasn't adjusted it. 

So, with the default 10-second wait, the 8th attempt would, exponentially, force a full restart, whereas you could have 90 retries with the 10-second constant wait in that same period.

But I could go either way.

DDD-59.md
```diff
@@ -0,0 +1,247 @@
...
+- Postgres already doesn't pin the extra pooled connections to the exported snapshot when
+  `snapshot.max.threads > 1`, so multi-threaded snapshots have a cross-connection consistency gap
+  today, independent of retry. Do we close that as part of this work (re-pin every pooled connection),
+  or keep this DDD scoped to retry and track the gap separately?
```
Naros: For this, I would advocate this is a bug and it should ideally be fixed in `main` now and backported to 3.6. This way that fix isn't prohibited by the discussion and turnaround of the DDD. WDYT?

Naros (commented):
This is a great write up @TimurRakhmatullin86, i've left some comments.

TimurRakhmatullin86:
Thanks @Naros — replies to each point, and I'll push a DDD update to match.

**Other connectors (MariaDB / SQL Server / Db2):** will expand the per-connector consistency section. MariaDB mirrors MySQL; SQL Server pins via snapshot isolation; Db2 I'll confirm from its snapshot source before asserting rather than guess.

**Oracle property alias:** agreed — I'll wire `snapshot.database.errors.max.retries` as a deprecated alias of the common property via the `Field` alias mechanism, so existing Oracle configs keep working and get the deprecation warning, then drop it in a release or two.

**Resume-from-key (your :241 idea):** I really like this. It turns retry from "re-emit the chunk" into "resume from the last-sent key," which kills both the duplicate rows and the double `FIRST`/`LAST` marker in one move. It fits chunked snapshots naturally since they already page by PK range — a retry just resumes from the high-water key; legacy single-table-per-thread gets an `ORDER BY` on the key columns plus the same high-water tracking. Two things to note, which I think are the source-timestamp provenance you meant: resumed rows are read in the retry's transaction so they carry that read's `ts_ms` (correct — they weren't emitted before — but a single chunk can then span two source reads); and it needs a stable, orderable key, so keyless tables fall back to the current re-read path. I'll restructure the Duplicates section around resume-from-key as the primary behavior, with duplicate-accept only as the keyless fallback.

**Retry budget:** agreed — one whole-snapshot budget, and per-unit granularity can come later if anyone needs it. Dropping the per-unit option.

**Backoff:** the Oracle snapshot-window math is convincing — constant as the default so the retries stay inside a bounded window. Exponential can be opt-in later with a max-delay cap so it can't blow the window.

**Postgres pinning (:247):** agreed it should be decoupled. I'll open a separate PR against `main` propagating the exported-snapshot pin to the extra pooled connections when `snapshot.max.threads > 1`, backportable to 3.6, and reference this DDD from it so the DDD can assume it's fixed.

Pushing the DDD update shortly.

hagar-yasser1:
Hello @TimurRakhmatullin86 and @Naros, 

I want to add to the discussion here the point of whether to enforce consistency across a snapshot or tradeoff snapshot consistency for the benefit of making the snapshot more resilient and lightweight. Relaxing the snapshot consistency constraint can be achieved by making each query to the DB in a new transaction with the proper isolation level. So the first trial for the snapshot may use a transaction with snapshot id `X` and if it fails the retry uses a transaction with snapshot id `X+1`. Eventual consistency of a row state can be reached after starting the cdc streaming. 

I would argue that this is very similar to incremental snapshots, however, in this case the cdc streaming will start after the snapshot. This is expected to cause out of order events when cdc starts (cdc will stream changes that has happened in the past relative to the snapshot) and maybe duplicates in case of a retry of part of the snapshot. 

I would also argue duplicate events is not a new tradeoff here, actually in a scenario where we would either retry a whole snapshot or only a chunk of it, this strategy will reduce the duplicate events that will get emitted to reach a finally successful snapshot. 

I understand that out of order events in a snapshot may be unexpected for some consumer types, however, I would argue it is not something totally strange. Please correct me if I am wrong, I think currently the consumers may get out of order events normally incase the connector restarts before committing the last read offset. The order of events can be always be checked by consumers through fields for log ordering like `LSN` for postgres events or `binlog file and position` for MySQL. 

Despite of that, I don't think the default behaviour should force an inconsistent snapshotting strategy, but is could be an option for consuming that can handle out of order events for a more resilient and flexible snapshotting strategy.

TimurRakhmatullin86:
@hagar-yasser1 good point, and I think it's a genuinely separate axis from what this DDD currently proposes — worth keeping distinct.

The DDD as written (resume-from-key) keeps the snapshot *consistent* and just avoids re-emitting rows on retry. What you're describing relaxes snapshot consistency itself: each (re)try reads its own fresh snapshot and streaming reconciles afterwards. You're right that it's close to how incremental snapshots already behave, and that consumers ordering by `LSN`/GTID — which already tolerate the out-of-order window a restart-before-offset-commit produces — could accept it.

I agree it shouldn't be the default. Where I'd land: this is an opt-in *snapshot consistency* mode (e.g. `snapshot.consistency = exported | relaxed`), orthogonal to the retry mechanism here — the retry loop is identical, only the "does each attempt reuse the pinned snapshot or take a fresh one" decision changes. That keeps the consistent path as the default and offers the resilient path to consumers that can handle out-of-order / duplicate events.

Whether it belongs in this DDD or a follow-up is really @Naros's call on scope. Happy to add a short section describing it as an optional mode if you both think it fits here.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+## Proposed changes
+
+### Configuration
+
+| Property | Type | Default | Meaning |
+|---|---|---|---|
+| `snapshot.retries.max` | int | `0` | Times the snapshot may retry a retriable table/chunk failure, as one **whole-snapshot** budget. `0` = current behavior. |
+| `snapshot.retry.delay.ms` | long | `10000` | Delay between attempts. |
+
+Per @Naros, this is a single whole-snapshot budget rather than per-unit — simpler contract, and
```
Naros: We can forgo the formality of attribution to reviewer names in the DDD. This seems to be the case in several places. While I appreciate it 😁 , it doesn't add any meaningful value.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
+# DDD-59: Generalizing per-table / per-chunk snapshot retry
```
Naros, line 1: Lets be sure that DDD-59 is added to the README.md file in the root of the repo.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+
+Tracking issue: [debezium/dbz#2297](https://github.com/debezium/dbz/issues/2297).
+Incorporates review feedback from @Naros on PR #60 (resume-from-key, whole-snapshot budget,
+constant backoff default, deprecated-alias for the Oracle property, Postgres pinning split out).
```
Naros: This is unnecessary at this point, the PR review process is more than sufficient to track this level of detail.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Motivation
+
+During the initial snapshot Debezium reads every captured table through a fixed-size thread pool,
```
Naros: Could you please make sure that each sentence is one line rather than wrapped, as it is currently? It makes suggestions and edit diffs much cleaner to review and understand.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+splitting large tables into chunks when configured
+(`RelationalSnapshotChangeEventSource#createDataEvents`). When one table or chunk read throws a
+`SQLException` — a connection reset partway through a large table, for instance — the callable wraps
+it in a `ConnectException`, it escapes the `ExecutorCompletionService`, and `doExecute` fails the
+whole snapshot.
+
+The only recovery today is the connector-level `errors.max.retries`, which restarts the snapshot
+from the beginning. On a large dataset that is the wrong unit of recovery: it re-reads the tables
```
Naros: It isn't necessarily the "wrong unit of recovery"; rather, it imposes the idea of full consistency without any real user flexibility for eventual consistency, which I would argue is the motivation here.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+offset and the consistent-read position were established. A chunk that fails for a transient reason
+should be recoverable by retrying *that chunk*.
+
+## Prior art
+
+The Oracle connector already does this connector-locally: `OracleSnapshotChangeEventSource` overrides
+`createDataEventsForTableCallable` with a bounded retry loop gated by
+`snapshot.database.errors.max.retries` (`int`, default `0`), with a narrow retriability check
+(`isTableSnapshotErrorRetriable` — only ORA-01466), backing off with `Metronome`, firing
+`notifyCompletedTableWithError` per attempt, hand-rolled rather than via `createPooledResourceCallable`.
```
Naros: It might be useful to outline here why a connector like Oracle gets away with this easily. This is because the snapshot phase uses Flashback queries, which allow the connector to read the table consistently across N threads as long as each thread uses the same SCN reference. No expensive lock, transaction, or other resource is necessary.

You could also link to the section "consistency per connector" where you describe this, too.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+- **MySQL / MariaDB** (MariaDB extends the same `BinlogSnapshotChangeEventSource`) hold a global read
+  lock and `START TRANSACTION WITH CONSISTENT SNAPSHOT` at REPEATABLE READ for the whole read phase,
+  released only after `awaitCompletion()`. No drift; the cost is availability (below).
+- **SQL Server** pins via `SNAPSHOT` isolation (`TRANSACTION_SNAPSHOT`) set at transaction start in
+  `connectionCreated`. A resumed read must re-establish the same snapshot isolation on a fresh
+  transaction.
+- **Postgres** uses the replication slot's exported snapshot (`SET TRANSACTION SNAPSHOT`), which today
+  is only pinned on the main connection — see the separate fix below.
+- **Db2** — snapshot mechanism to be documented from the `debezium-connector-db2` repo (not verified
+  here); left as a follow-up rather than asserted.
```
Naros: Have you had a chance to review Db2? There is also CockroachDB, Ingres, Informix, IBMi, and YashanDB that should also likely be included so we have a clear broad spectrum of insights.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+per-unit granularity can be added later if needed (easier than walking it back). Following house
+style (`SNAPSHOT_LOCK_TIMEOUT_MS` / `snapshotLockTimeout()` returning `Duration`), the delay getter
+returns a `Duration` and the max field is `Field::isNonNegativeInteger`-validated.
+
+**Reconciling with Oracle.** The common property becomes canonical; Oracle's
+`snapshot.database.errors.max.retries` is wired as a deprecated alias via
+`Field.withDeprecatedAliases(...)` (the same mechanism Oracle already uses for
+`DEPRECATED_XSTREAM_SERVER_NAME`), so existing Oracle configs keep working and get the deprecation
+warning, and the alias can be dropped after a release or two. Oracle keeps only its retriability
+override (ORA-01466), not the whole loop.
```
Naros: Let's put this in a Backward compatibility section so that its clear that this is how we're handling existing connector configurations with any newly presented configuration options.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+across a connector restart (v1 keeps the high-water key in memory for same-run retry — see caveats).
+
+## Proposed changes
+
+### Configuration
+
+| Property | Type | Default | Meaning |
+|---|---|---|---|
+| `snapshot.retries.max` | int | `0` | Times the snapshot may retry a retriable table/chunk failure, as one **whole-snapshot** budget. `0` = current behavior. |
+| `snapshot.retry.delay.ms` | long | `10000` | Delay between attempts. |
```
Naros: These two properties have unique stems: `snapshot.retries` vs `snapshot.retry`. We prefer that settings categorized to work together, as is the case here, reside in the same logical stem to be consistent with our existing conventions. 

I'm not partial to either "retries" or "retry", but would prefer they be the same.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+8. Docs: the properties, the keyless duplicate-accept fallback, the `ts_ms` note, and per-connector
+   consistency.
+
+## Open questions
+
+- Cross-restart resume (persisting the last-emitted key in the offset) — v1 or follow-up? Proposed
+  follow-up.
+- Db2 snapshot-consistency mechanism — needs the `debezium-connector-db2` repo to document.
+- Whether the legacy-path `ORDER BY <key>` should be unconditional or only when retries are enabled
+  (to avoid the Postgres sort cost for users who don't opt in).
```
Naros: By "legacy path", I assume you mean single-threaded snapshots? If so, yes, I would say that in practice, if snapshot retry isn't opted in, no connector should explicitly specify an `ORDER BY` clause and instead prefer the primary key index sort by default to allow for optimal performance.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+   marker; fails N+1 → snapshot fails with the attempt count; `0` reproduces current behavior; a
+   non-retriable error fails immediately without consuming the budget; `InterruptedException` aborts
+   regardless; keyless table falls back to re-read.
+8. Docs: the properties, the keyless duplicate-accept fallback, the `ts_ms` note, and per-connector
+   consistency.
+
+## Open questions
+
+- Cross-restart resume (persisting the last-emitted key in the offset) — v1 or follow-up? Proposed
+  follow-up.
```
Naros: I wonder if it's sufficient for initial snapshots for them to be less susceptible to network hiccups in v1, but if the maximum number of retries is exhausted and the connector falls back to the runtime, we don't have a way to resume.

I may be in the minority here, but cross-restart resuming of snapshots is something that I think is best left to things like ad-hoc blocking or incremental snapshots, allowing us to build on that existing foundation and offset management tier.

DDD-59.md
```diff
@@ -0,0 +1,220 @@
...
+(Oracle ~15 min OOTB), exponential backoff from a 10s base forces a full restart by roughly the 8th
+attempt, whereas constant 10s allows ~90 retries in the same window. Exponential can be an opt-in
+later with a max-delay cap so it can't blow a connector's window.
+
+### Observability
+
+Snapshot metrics run through `SnapshotProgressListener` → `SnapshotMeter` → the `SnapshotMetricsMXBean`
+trait, and existing per-table metrics expose `Map<String, Long>` breakdowns. A retry counter matches
+that shape: `Map<String, Long> getTableSnapshotRetries()` on the trait, a listener callback, and a
+`ConcurrentMap` field/mutator on `SnapshotMeter`, plus a WARN log per retry.
```
Naros: I wonder if we need a metric here or if perhaps tracking this via the log entry or a notification event, similar to what we do when chunks or tables are completed in the snapshot process, is sufficient.

Naros:
> Where I'd land: this is an opt-in snapshot consistency mode (e.g. snapshot.consistency = exported | relaxed), orthogonal to the retry mechanism here — the retry loop is identical, only the "does each attempt reuse the pinned snapshot or take a fresh one" decision changes. That keeps the consistent path as the default and offers the resilient path to consumers that can handle out-of-order / duplicate events.

Do we need to get this far into the weeds?

Is it reasonable enough to document when setting `snapshot.max.threads` with a value greater than 1 that certain databases X, Y, and Z can provide consistent snapshots across multiple threads, whereas databases A, B, and C are only eventually consistent? 

It then becomes a user-driven choice, just like the proposed `snapshot.consistency` option but without adding new configs. This also keeps our hands clean from needing to try and develop an "exported" or "consistent" snapshot mechanism for a database that truly is incapable of supporting that across 2+ threads.

@jpechane what's your take?

DDD-59.md
TimurRakhmatullin86: Good call — dropped the reviewer @-mentions throughout the doc.

DDD-59.md
TimurRakhmatullin86: Added — DDD-59 is now listed in `README.md` in this push.

DDD-59.md
TimurRakhmatullin86: Removed the changelog/feedback block from the header — agreed the PR history tracks this well enough.

DDD-59.md
TimurRakhmatullin86: Done — reflowed the whole document to one sentence per line.

DDD-59.md
TimurRakhmatullin86: Reworked the Motivation around exactly this. The point now is that the whole-snapshot restart forces full re-consistency with no way to opt into eventual consistency — re-read just the failed unit and let consumers dedupe on key — and that flexibility is the real motivation. "Wrong unit of recovery" is gone.

DDD-59.md
TimurRakhmatullin86: Added to Prior art: Oracle bakes the SCN into each per-table Flashback query (`… AS OF SCN`), so every thread reads at the same SCN with no shared lock, transaction, or resource to preserve across a retry — which is why its per-table retry needs no extra machinery. Linked forward to the per-connector consistency section.

DDD-59.md
TimurRakhmatullin86: Broadened it, and reorganized the section into a three-class taxonomy so the pattern is visible rather than a flat list:

- **Class A — the pin travels in the query text, parallel-safe for free:** Oracle (`AS OF SCN`, verified in-repo).
- **Class B — a shareable exported snapshot that isn't shared yet, fixable to safe:** Postgres (`SET TRANSACTION SNAPSHOT`, verified; this is the #7740 fix).
- **Class C — txn/connection-bound isolation with no shareable view, drift-prone under multi-connection parallelism:** MySQL/MariaDB (kept mutually consistent only by the global read lock, not a shared snapshot) and SQL Server (verified).

For the connectors you listed — Db2 LUW, Db2 for i (IBMi), Informix, Ingres, CockroachDB, YashanDB — I want to be upfront: none of them are in this repository (they're separate community repos), so I've added them as **inferred** rows drawn from each database's documented snapshot semantics and labeled them clearly as *not verified against connector source*. My best reading: CockroachDB (`AS OF SYSTEM TIME`) and YashanDB (Oracle-compatible `AS OF SCN`) would be Class A if implemented that way; Db2/IBMi/Informix/Ingres are isolation-level based with no exported snapshot, so Class C alongside SQL Server. I've noted that these should be confirmed against their real source when those connectors move onto this common path — happy to do that if you can point me at the repos.

DDD-59.md
TimurRakhmatullin86: Done — moved the deprecated-alias handling into its own Backward compatibility section.

DDD-59.md
TimurRakhmatullin86: Unified to a single stem: `snapshot.retry.max` and `snapshot.retry.delay.ms`.

DDD-59.md
TimurRakhmatullin86: Yes — "legacy path" means the single-threaded snapshot (`snapshot.max.threads = 1`). Updated so it stays exactly as-is (no `ORDER BY`) when retry isn't opted in; only when retry is enabled do we sort by the primary key to make resume deterministic — the index's natural order (free) on clustered-PK engines, an accepted cost on a Postgres heap / large secondary-PK table. Dropped the matching open question since this is now decided.

DDD-59.md
TimurRakhmatullin86: Agreed — dropped cross-restart resume from v1 as out of scope. If the retry budget is exhausted the connector falls back to the runtime error path exactly as today, and a snapshot that needs to resume across a restart is already served by ad-hoc blocking / incremental snapshots, which persist their own state. v1 retry is same-run only, by design rather than as a stopgap.

DDD-59.md
TimurRakhmatullin86: Switched — no new metric. Retries are surfaced by a WARN log per attempt plus a notification through the existing `NotificationService`, matching how table/chunk completion is already reported. Left a note that a counter can be added later if one turns out to be wanted.

TimurRakhmatullin86:
Pushed a revision addressing all the comments above — replied inline on each thread. Quick summary of what changed:

- **Motivation** reframed around eventual-consistency flexibility (not "wrong unit of recovery").
- **Prior art** now explains *why* Oracle's per-table retry is cheap (Flashback `AS OF SCN`), linked to the consistency section.
- **Consistency per connector** reorganized into a three-class taxonomy (pin-in-query-text / shareable-exported-snapshot / txn-bound) and broadened to Db2, Db2 for i, Informix, Ingres, CockroachDB, YashanDB.
- **Backward compatibility** split into its own section; config unified to the `snapshot.retry` stem; `ORDER BY` on the single-threaded path made conditional on opt-in; cross-restart resume dropped as out of scope; retries surfaced via log + notification instead of a metric; reviewer @-mentions and the changelog header removed; DDD-59 added to `README.md`; whole doc reflowed to one sentence per line.

One thing I want to flag rather than bury: of the connectors you listed, **only Oracle/MySQL/MariaDB/SQL Server/Postgres are in this repo** — Db2, Db2 for i, Informix, Ingres, CockroachDB, and YashanDB live in separate community repos, so I've added them as clearly-labeled **inferred** rows from each database's documented snapshot semantics, not verified against connector source. If any of those readings are off I'd rather correct them against the real code — pointers welcome.

hagar-yasser1:
Adding to @Naros's point here:

> Is it reasonable enough to document when setting snapshot.max.threads with a value greater than 1 that certain databases X, Y, and Z can provide consistent snapshots across multiple threads, whereas databases A, B, and C are only eventually consistent?

I would say that the leverage of switching off the consistency guarantee for databases like Postgres is still useful. (where we can either force a snapshot number to guarantee consistency or relax the consistency guarantee). 
I can understand that one consumer may prefer to keep the snapshots consistent if they can and another consumer would prefer to relax the consistency guarantee consciously to reduce the MVCC load on the source DB and to increase the chance of making the snapshot successful. 

Please correct me if I am wrong, as I understand in the case where a user forces consistency in an Oracle or Postgres snapshot while enabling retries, there is a chance that if a chunk fails and we retry opening a transaction from the same snapshot number that this version of the data is already purged by the database management system. In that case all the retries will fail and there will be no way to take a successful snapshot except retrying the whole snapshot in the beginning. This is an example of when a consistency switch off is helpful if the DB provides the option for both modes.

TimurRakhmatullin86:
Great point, and I think it's exactly right — it's the strongest argument for the relaxed mode being a first-class option rather than an afterthought.

Two things worth separating, because the failure you describe behaves differently per connector:

- **Oracle** is precisely your scenario. A retry re-runs `… AS OF SCN <scn>`, and once the undo for that SCN has been overwritten (past `UNDO_RETENTION`), the read fails with **ORA-01555 "snapshot too old."** From that point every retry at that SCN fails identically, so per-chunk retry can't recover — only a whole-snapshot restart at a fresh SCN can. The backoff window actually makes this *more* likely, since it's more wall-clock time for the undo to age out.
- **Postgres** gets there by a different route. While a transaction still holds the exported snapshot, the tuples it needs are protected from vacuum (the global xmin is held back), so you don't hit "purged data" on read the way Oracle does. What you trade instead is a long-lived transaction — xmin-horizon pressure and bloat on the source — and the hard failure is if the *exporting* transaction is lost, after which `SET TRANSACTION SNAPSHOT '<name>'` can no longer be re-established at all.

So the design line I'd draw from this: **per-chunk retry has a ceiling in consistent mode.** It's the right tool for the transient/connection-reset class (dbz#2297), but it cannot resurrect a point-in-time the source has already aged out. That gives a clean division of responsibility rather than a conflict:

1. per-chunk retry (this DDD) → transient failures within a still-valid snapshot;
2. the existing whole-snapshot `errors.max.retries` → the backstop for "the pinned point is gone," where only a fresh restart helps;
3. an opt-in **relaxed/eventual** mode → the escape hatch that lets a fresh per-chunk read at a *new* point-in-time succeed where a pinned one can't — which is exactly the flexibility the Motivation argues for, and your MVCC-load point is a second independent reason to want it.

On @Naros's documentation question (which databases give consistent-across-threads vs only eventual under `snapshot.max.threads > 1`): the revised **Consistency per connector** section now answers that as a three-class split — Class A (Oracle; and, if implemented that way, CockroachDB/YashanDB) is consistent across threads for free; Class C (SQL Server, Db2, Informix, Ingres, IBMi — and MySQL/MariaDB except while the global read lock is held) is only eventual unless you pin to one thread. I'll add a short explicit note tying that table to the two retry modes and to this aging-out ceiling, so it's stated outright rather than left implied — does that framing match what you both had in mind?

hagar-yasser1:
Hello @TimurRakhmatullin86 and @Naros ,

How do you think we should proceed here?

TimurRakhmatullin86:
Hi @hagar-yasser1, thanks for the nudge. Here's a concrete path I'd suggest to get this moving:

**Where we landed on the design.** The discussion converged on a clean division of responsibility rather than one knob trying to do everything:

1. **Per-chunk retry (this DDD)** — recovers *transient* failures (connection resets, dbz#2297) within a still-valid snapshot, resuming from the last-emitted key so it doesn't duplicate rows.
2. **The existing whole-snapshot `errors.max.retries`** — stays as the backstop for the "pinned point-in-time is gone" case (Oracle ORA-01555, a lost Postgres exporting transaction), which per-chunk retry provably cannot recover.
3. **An opt-in relaxed/eventual mode** — the escape hatch that lets a fresh read at a *new* point-in-time succeed where a pinned one can't, and reduces MVCC load; consumers dedupe on key.

**Proposed next steps:**
- I'll fold the "consistency modes + aging-out ceiling" outcome into the DDD as an explicit section (turning my earlier note into actual text), and scope **v1 = per-chunk retry for transient failures** with the whole-snapshot backstop unchanged, leaving the relaxed/eventual mode as a fast-follow so v1 stays small and reviewable.
- Once @Naros is happy with that scope, I'm glad to open the implementation PR(s) against `debezium/debezium` — the common retry loop + resume-from-key in `RelationalSnapshotChangeEventSource`, then per-connector retriability predicates.

@Naros — does that v1 scope and the three-way division match your thinking? If so I'll update the doc accordingly and start on the implementation. Happy to split the work with @hagar-yasser1 if you'd like to collaborate on it.

Naros:
Hi @TimurRakhmatullin86 and @hagar-yasser1 see inline

> * per-chunk retry (this DDD) → transient failures within a still-valid snapshot;
> * the existing whole-snapshot `errors.max.retries` → the backstop for "the pinned point is gone," where only a fresh restart helps;

These two make sense to me.

> * an opt-in **relaxed/eventual** mode → the escape hatch that lets a fresh per-chunk read at a _new_ point-in-time succeed where a pinned one can't — which is exactly the flexibility the Motivation argues for, and your MVCC-load point is a second independent reason to want it.

I'm still not entirely sold on this one. I think if we were to consider this as an option, it would need to be documented well and similarly to how we document snapshot locking/isolation modes: that if you relax the knobs, the snapshot can fail if there are schema changes, and that's not a bug. The user takes ownership in guaranteeing no DDL happens during the snapshot when using this mode.

> I'll fold the "consistency modes + aging-out ceiling" outcome into the DDD as an explicit section (turning my earlier note into actual text), and scope v1 = per-chunk retry for transient failures with the whole-snapshot backstop unchanged, leaving the relaxed/eventual mode as a fast-follow so v1 stays small and reviewable.

I definitely think this is a good scoped goal, fine with me :rocket:

TimurRakhmatullin86:
Thanks @Naros — that scope works for me, I'll run with it. 🚀

**v1** (this PR series): per-chunk/per-table retry for *transient* failures with resume-from-key, and the existing whole-snapshot `errors.max.retries` left untouched as the backstop for the "pinned point is gone" case. I'll fold the "consistency modes + aging-out ceiling" outcome into the DDD as an explicit section so the boundary between the two is written down.

**Relaxed/eventual mode** — agreed, and I'll keep it out of v1 as a fast-follow. When it does land, I'll document it the same way as the snapshot locking/isolation modes: enabling it trades the consistency guarantee, so a concurrent DDL/schema change during the snapshot can make it fail, and that is expected rather than a bug — the operator takes ownership of ensuring no DDL runs during the snapshot in that mode. That framing (a knob you opt into, with the caveat spelled out) matches how the existing modes are presented, so it shouldn't feel bolted on.

I'll start on the v1 implementation PR against `debezium/debezium` (the common retry loop + resume-from-key in `RelationalSnapshotChangeEventSource`, with a connector-overridable retriability predicate defaulting to no-op) and link it here. Thanks for the reviews on this — it's in a much better shape than where it started.

DDD-59.md
hagar-yasser1: Hello @TimurRakhmatullin86 and @Naros, I missed this point in the DDD and just noticed it in the [PR](https://github.com/debezium/debezium/pull/7990). I think it is important to enable the user to set the maximum retries per chunk as well. For example, if I trigger a chunked snapshot with 2000 chunks, I would want to guarantee that if any chunk failed it would retry but not more than 3 times. I wouldn't say all chunks may fail so,  if I set a budget of 100 retries that may cause these retries to be exhausted quickly in a persisting failing chunk which I don't want to retry 100 times before the snapshot fails.

WDYT?

PR #70, opened 2026-08-31
DDD-68 Debezium Platform Snapshot Monitoring

mfvitale:
Provide high level design to enable Debezium Platform to monitor snapshots progress

### DIFF
--- a/DDD-68.md
+++ b/DDD-68.md
@@ -375,6 +375,16 @@ For incremental snapshots, per table chunk progress may not be available (it dep
 - Rows scanned counter updates as notifications arrive.
 - Global progress (tables completed out of total) still works regardless.
 
+#### How we intend to fill the gap
+
+To turn the incremental per table view from a counter into a real progress bar, the connector needs a per table total row count at snapshot start (initial snapshots already report chunk totals). That count is resolved by an automatic strategy with no configuration, chosen so we pay the expensive path only when we have to:
+
+1. If the snapshot has no additional filter condition and the database exposes a cheap row count estimate (for example Postgres `pg_class.reltuples` or the InnoDB row estimate), use that estimate. It is a constant time metadata read regardless of table size.
+2. Otherwise (a filter is present, or the database has no estimate source), run an exact count bounded to the rows the snapshot will actually scan, that is `count of rows with primary key up to the snapshot maximum key`, applying the same filter when one exists.
+3. If the count query fails for any reason, resolve to no count. The view then stays in the spinner plus counter mode described above.
+
+The bound on the primary key keeps the total on the same basis as the rows the snapshot really reads, so the bar cannot exceed 100 percent when new rows arrive during the snapshot. Because estimates can be slightly stale, the total is treated as best effort: the frontend clamps the bar and keeps the rows scanned counter accurate. From the frontend point of view nothing changes when this lands, the per table `progress` field simply starts being populated and the bar replaces the spinner.
+
 ## Snapshot Lifecycle and States
 
 The platform maps raw Debezium notifications into a simplified lifecycle. Both snapshot types share this model.
--- a/DDD-68.md
+++ b/DDD-68.md
@@ -1,0 +1,423 @@
+# DDD-68: Snapshot Monitoring for Debezium Platform
+
+## Motivation
+
+The platform can already trigger incremental snapshots (from the Actions tab) and runs an initial snapshot automatically on first pipeline start (driven by `snapshot.mode`). What it cannot do today is show what a snapshot is doing while it runs.
+
+Once a snapshot starts there is no visibility into its progress. Users cannot tell whether a snapshot is still running, which table is being processed, how many tables remain, or whether it has stalled. The only way to check is to read the connector logs, which is slow, requires cluster access, and is not something a platform operator should have to do for a routine operation.
+
+A snapshot of a large database can run for a long time. During that window downstream consumers are waiting for a complete backfill, and the operator has no way to set expectations or to notice that a single large table (or a permission error on one table) is holding everything up.
+
+This document proposes a snapshot monitoring feature that turns the raw progress signals Debezium already emits into a clear, real time view in the UI: a global progress bar, a per table breakdown, rows scanned counters, and a history of completed runs.
+
+## Goals
+
+Primary objectives:
+
+1. Give operators real time visibility into initial and incremental snapshot progress, per pipeline, without reading connector logs.
+2. Show both a global view (tables completed out of total) and a per table view (status, rows scanned, and chunk progress when available).
+3. Present initial and incremental snapshots through one uniform UI and one uniform API, despite their differences at the notification level.
+4. Persist completed snapshot runs as history so operators can review past durations, outcomes, and per table results.
+5. Push updates to the UI as they happen (no polling) while keeping the client simple.
+6. Survive restarts of both the Debezium Server and the Conductor without losing progress state.
+
+Non goals for this phase:
+
+- No global "all snapshots across all pipelines" page. Monitoring is scoped to one pipeline.
+- No control actions from this feature. It is read only. Triggering snapshots stays in the Actions tab.
+- No scan rate / throughput indicator, no ETA, and no snapshot alerting. These are listed under Future Enhancements.
+- No feature specific metrics/observability export in this phase.
+
+## Proposed Solution
+
+Debezium already emits `Notification` events during both initial and incremental snapshots (STARTED, IN_PROGRESS, per table completion, COMPLETED, and so on). The platform does not consume these yet.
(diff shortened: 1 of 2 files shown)


mfvitale:
@indraraj Could you take a look to the API and UI/UX?

DDD-68.md
```diff
@@ -0,0 +1,423 @@
...
+| `IDLE` | No snapshot running | Hide the Overview card; show "No active snapshot" plus history in the tab |
+| `RUNNING` | Actively scanning tables | Animated progress bar and table list |
+| `PAUSED` | Incremental snapshot paused | Static bar, "Paused" badge |
+| `COMPLETED` / `ABORTED` / `SKIPPED` / `UNKNOWN` | Terminal | The run moves to history; the live view returns to IDLE |
+
+The frontend does not need to interpret raw Debezium notifications or track transitions. It renders the latest full state and, for history, reads the REST endpoints.
+
+### D. Graceful degradation
+
+For incremental snapshots, per table chunk progress may not be available (it depends on an optional Debezium core enhancement). In that case `progress` is `null` and the UI shows a spinner and the rows scanned counter rather than a bar. Global progress (tables completed out of total) always works.
```
mfvitale, line 368: @jpechane @Naros Since for the incremental snapshot we do not have clear information about the total rows/ chunks, the idea is to improve the incremental snapshot information adding those info. This will require to run a `COUNT(*)` query queries after/before the `buildMaxPrimaryKeyQuery`. 

But, I have some concern about the performance implication. So, what is your thought? 

An option could be to use some estimation using metadata table like `pg_class.reltuples` for PostgreSQL.

Naros: I see two paths here, depending on `additionalConditions`. I would use a second query with these rules:

* If `additionalConditions` is empty and the database has metadata estimate tables, use them.
* If `additionalConditions` is not empty OR the database has no metadata estimate table, use a `COUNT(1)` hard select on the table, including the conditions if they're not empty.

This would give per connectors a chance to override the method based on the metadata estimates table existance accordingly. This way, you take the expensive route only when necessary, and the optimized route when possible. WDYT?

mfvitale: @Naros I like the approach.

DDD-68.md
```diff
@@ -0,0 +1,433 @@
...
+
+A snapshot of a large database can run for a long time. During that window downstream consumers are waiting for a complete backfill, and the operator has no way to set expectations or to notice that a single large table (or a permission error on one table) is holding everything up.
+
+This document proposes a snapshot monitoring feature that turns the raw progress signals Debezium already emits into a clear, real time view in the UI: a global progress bar, a per table breakdown, rows scanned counters, and a history of completed runs.
+
+## Goals
+
+Primary objectives:
+
+1. Give operators real time visibility into initial and incremental snapshot progress, per pipeline, without reading connector logs.
```
indraraj, line 17: Are we going to add this visibility to the pipeline listing page also, something that can be annotated (e.g _Initial snapshot 42%_) while a snapshot is in progress? Do you think this will be useful to the user?

DDD-68.md
```diff
@@ -0,0 +1,433 @@
...
+2. Show both a global view (tables completed out of total) and a per table view (status, rows scanned, and chunk progress when available).
+3. Present initial and incremental snapshots through one uniform UI and one uniform API, despite their differences at the notification level.
+4. Persist completed snapshot runs as history so operators can review past durations, outcomes, and per table results.
+5. Push updates to the UI as they happen (no polling) while keeping the client simple.
+6. Survive restarts of both the Debezium Server and the Conductor without losing progress state.
+
+Non goals for this phase:
+
+- No global "all snapshots across all pipelines" page. Monitoring is scoped to one pipeline.
+- No control actions from this feature. It is read only. Triggering snapshots stays in the Actions tab.
```
indraraj, line 27: On a side note: In the Pipeline > Actions tab, there is currently no check that the source has `signal.data.collection` configured before triggering an incremental / blocking snapshot. Since we already have the source config here, we could check if its missing and warn the user here and in case its not present maybe add a UI shortcut to configure signalling on the source. wdyt?

DDD-68.md
```diff
@@ -0,0 +1,433 @@
...
+
+- Show snapshot type, a global progress bar, the percentage, and the table count (`completedTables`/`totalTables`).
+- Show elapsed time. Do not show a scan rate; that is intentionally out of scope.
+- When `status` is `PAUSED`, show a "Snapshot Paused" heading and a "Paused" badge, and a static (non animated) bar with a "Paused after Xm Ys" line.
+- "View Details" navigates to the Snapshots tab.
+- Data source: the one shot `/progress` endpoint for first paint, then the SSE stream for live updates.
+
+### B. Snapshots tab (dedicated)
+
+This is a new tab, separate from the Monitoring tab (which hosts the Prometheus observability panels). The tab shows an active state badge while a snapshot is running.
```
indraraj, line 268: From the Actions tab, users can also trigger a blocking snapshot. And here we're planning to support initial and incremental snapshots. Calling the tab just "Snapshot" might lead users to expect the blocking snapshot to be covered as well.

DDD-68.md
```diff
@@ -0,0 +1,433 @@
...
+│  Snapshot Progress                                                    │
+├─────────────────────────────────────────────────────────────────────┤
+│                                                                      │
+│  Initial Snapshot   ██████████████░░░░░░░░  58%                      │
+│  Status: Running   |   Started: 10:00:12 UTC   |   Elapsed: 5m 32s  │
+│  Tables: 7 of 12 completed   |   Total rows scanned: 245,000        │
+│                                                                      │
+│  ── Tables ─────────────────────────────────────────────────────     │
+│                                                                      │
+│  Table                     │ Status       │ Progress       │ Rows    │
```
indraraj, line 283: Any idea what the expected max table count would be for a larger edge case?

DDD-68.md
```diff
@@ -0,0 +1,433 @@
...
+
+## Frontend / UI Requirements
+
+The feature appears in two places on the pipeline detail page.
+
+### A. Overview tab: compact progress card
+
+Shown only when a snapshot is active for the pipeline. Hidden when IDLE.
+
+```
```
indraraj, line 238: I guess multiple cards in case of more than one active snapshot at a time? Or can it only have one at a time?

DDD-68.md
indraraj: @mfvitale here

mfvitale: Why not? Could you propose an UX/UI?

DDD-68.md
mfvitale: Let's wait how [this](https://github.com/debezium/debezium/pull/8025) ends up and we can then evaluate to integrate it. WDYT?

DDD-68.md
mfvitale: It is covered since effectively it is an on demand initial snapshot. It uses the same notifications.

DDD-68.md
mfvitale: Millions/Billions? @Naros Worst thing you saw in your experience?

DDD-68.md
mfvitale: Only one at time.

indraraj: So, does this mean we can't trigger another snapshot until the previous one is completed, or would it be queued and run in sequence? If we can't trigger more than one at a time, we should add this check in the Actions tab UI.

mfvitale: You can trigger another one but the result is that the new tables added to the new one will be effectively added to the current on list of table to snapshot.

DDD-68.md
indraraj: here https://github.com/mfvitale/debezium-design-documents/pull/1

DDD-68.md
indraraj: With an edge case this large, I can easily see us getting into thousands of table entries quite quickly. In the Snapshots tab, the SSE event currently carries the full tables[] array, and for a large number of tables, this could potentially freeze the UI tab. A virtualised list could help initially, but for a much larger number of tables (multi-thousands), it may still have limitations since the UI would still need to hold every table object in memory if the event includes them all. 

Maybe we'll need to look into a more dedicated solution later. Could we bound what the live event sends and load the rest on demand?

DDD-68.md
indraraj: sure

DDD-68.md
mfvitale: Oh sorry I wrongly read. You meant table not row. Then I think we are more around thousands.

DDD-68.md
mfvitale: @indraraj Could you incorporate the changes?

indraraj (approved):

PR #76, opened 2026-09-16
debezium/dbz#74 Add DDD for Graph-to-Relational Debezium SMT for Neo4j

### LINKED ISSUE #74: DDD for Neo4j SMT graph-to-relational

vsantonastaso:
Closes: DDD-74 (#74)
 
## Summary

This DDD proposes a Kafka Connect SMT that transforms Neo4j's event to Debezium format.. 
The DDD focuses on the **Graph -> Relational** direction (V1). 
The reverse direction (Relational -> Graph) is deferred to a  DDD-55 (#55).


DDD-74.md
```diff
@@ -0,0 +1,517 @@
...
+The mapping quality depends on how the Neo4j source connector is configured. V1 assumes the CDC source strategy, `EXTENDED` payload mode, and key/uniqueness constraints on every mapped entity. These are described in [Required Neo4j source connector settings](#required-neo4j-source-connector-settings).
+
+#### Operation Mapping
+
+Neo4j CDC operations map to Debezium operations one-to-one:
+
+| Neo4j `event.operation` | Meaning | Debezium `op` | Row image used |
+|:---|:---|:---|:---|
+| `c` | Create | `c` | `state.after` |
+| `u` | Update | `u` | `state.after` (and `state.before` for the key) |
```
Naros, line 78: Since this is going through the SMT and we can't emit two records, should a `u` with changing key values be an error condition? 

In relational sources, if a key column value drifts, this triggers the `u` to be rewritten as a `d` followed by a `c` event. This is the expected form the JDBC sink expects, and if the SMT doesn't provide it, the row would be treated as an insert rather than an update.

DDD-74.md
```diff
@@ -0,0 +1,517 @@
...
+| `relationship.<TYPE>.properties` | String | No | All | Relationship properties to include as columns |
+| `relationship.<TYPE>.owner` | String | No | `start` | (foreign_key) Which endpoint owns the row to update (`start` or `end`) |
+| `relationship.<TYPE>.fk.column` | String | No | Derived from the non-owner label + key | (foreign_key) Foreign-key column populated from the non-owner endpoint key |
+
+#### Global configuration
+
+| Property | Type | Required | Default | Description |
+|:---|:---|:---|:---|:---|
+| `field.missing.behavior` | String | No | `warn` | How to react when a required key or image is missing: `fail`, `warn` (drop record), or `ignore` (drop silently). Mirrors the Neo4j CUD Converter |
+| `tombstones.enabled` | Boolean | No | `true` | Emit a tombstone after a delete for log compaction |
```
Naros: Does this mean whether the `tombstone` from the Kafka topic is passed through the SMT or dropped? The description gives the impression that the SMT emits a tombstone after a `d`, which I don't see how is possible, given that the SMT contract is limited to just 1 event.

DDD-74.md
```diff
@@ -0,0 +1,517 @@
...
+3. disambiguate a **multi-label** node (which label owns the row);
+4. handle a label or relationship that has **no constraint**, so `event.keys` is empty.
+
+The rest of this section details those overrides: naming first, then the per-entity and global property tables.
+
+#### Naming overrides
+
+Table and column names do not usually need configuring: the SMT derives them from the event and applies them itself, setting the output topic to the target table (which the JDBC sink writes to via `collection.name.format=${topic}`) and naming the columns from the properties. Overrides are needed only when the target database already exists with different names. Two cases:
+
+**Greenfield target → no naming configuration needed.** If the target database is empty and the Debezium JDBC sink creates the tables (`schema.evolution=basic`), the convention names are used with the default `table.naming=as_is`: a `Customer` node becomes a `Customer` table keyed by `id`, and a `CONTAINS` relationship becomes a `CONTAINS` table with columns `order_id`, `product_id`, `quantity`. Nothing external dictates other names, so there is nothing to override.
```
Naros: There are a few places where we show `Order` and `id` but then report `order_id` as the column. I don't recall seeing a case specification rule; is this just an oversight?

DDD-74.md
```diff
@@ -0,0 +1,517 @@
...
+{
+  "before": null,
+  "after": { "id": 5001, "customer_id": 1004 },
+  "source": { "connector": "neo4j", "table": "orders", "txId": 16 },
+  "op": "u",
+  "ts_ms": 1677844800000
+}
+```
+
+This is a **partial update**: `after` carries only the owner's key (`id`) plus the foreign-key column (`customer_id`), so the Debezium JDBC sink upsert leaves the other `orders` columns untouched when the row already exists.
```
Naros, line 258: For partial updates, does Neo4j intend to show only that the FK relationship changed? If some other field that isn't the FK had changed, would there be a full record or would it also be a partial?

Naros (commented):
Hi @vsantonastaso I've left some questions below.

DDD-74.md
vsantonastaso: Yes, you're right. I missed this case. I don't think we can handle it in a simple stateless SMT and if I treat it as an insert for the new key, the old one becomes orphaned. So, I'll add it to the constraints and limitations section.

DDD-74.md
vsantonastaso: Yep, my fault. It's not possible to emit multiple events, so I'll treat it as I did in Neo4jCudConverter, when enabled, I'll pass the event, otherwise, I'll drop it.

DDD-74.md
vsantonastaso: Yes, with the `column.naming=as_is` prop the column will be called 'Order_id'. To force lowercase, we should use the `column.naming=snake_case` property. It is a global configuration setting, so it will affect all columns, but I don't think it's worth adding a specific property only for the join table. I can't think of a use case where the user would want everything to be in the _as_is_ format except for the join table columns. WDYT?

DDD-74.md
vsantonastaso: yep right, the relationship events (eventType="r") and node events (eventType="n") are separate. A relationship change emits only endpoint keys , so always a partial FK update. A change to any non-FK node field is a separate node event carrying full state.after.properties, so full-row upsert. I'd add a sentence about it

vsantonastaso:
> Hi @vsantonastaso I've left some questions below.

Thanks for the review @Naros . I tried to answer and address your comments.

DDD-74.md
```diff
@@ -0,0 +1,542 @@
...
+## Out of scope
+
+- **Relational-to-Graph SMT** (the opposite direction): handled by the Neo4j CUD Converter.
+- **Stateful operations**: cross-topic key resolution, relationship-to-foreign-key back-patching that needs the full target row, and schema reconciliation across events all require state and are out of scope.
+- **Neo4j schema/index change events**: Neo4j CDC does not emit schema or index changes, so there is nothing to propagate as DDL. Constraint and index management on both sides is the user's responsibility.
+- **Query-based source strategy**: V1 assumes the Neo4j source connector runs with the CDC source strategy, which produces structured `before`/`after` change events. The custom-query strategy is out of scope.
+
+## Architecture
+
+The SMT is deployed on the **Debezium JDBC sink connector** side.
```
mfvitale:
```suggestion
The SMT is configured on the **Debezium JDBC sink connector** side. 
```

DDD-74.md
```diff
@@ -0,0 +1,542 @@
...
+The `keys` field is a map keyed by label, each value a list of key-property maps:
+
+```json
+"keys": {
+  "Customer": [ { "id": 1004 } ]
+}
+```
+
+If no constraint exists, `keys` is empty and the SMT cannot derive a portable primary key. 
+V1 therefore **requires the relevant Neo4j constraints to exist** and treats a missing key as a `field.missing.behavior` condition (fail/warn/ignore).
```
mfvitale, line 63: Just a question. Could the missing of `keys` be managed with a ~`MERGE/MATCH`~ an hash from all properties?

DDD-74.md
```diff
@@ -0,0 +1,542 @@
...
+
+| Neo4j field | Debezium envelope | Description |
+|:---|:---|:---|
+| matched `label` | **output topic** → target table | `label.<Label>.table` if set, else the label transformed by `table.naming` |
+| `keys.<Label>` entry | `after`/`before` key columns + record key | Primary-key columns |
+| `state.after.properties` | `after` non-key columns | Row columns |
+| `state.before.properties` | `before` (for `u`/`d`) | Previous row image |
+| `event.operation` | `op` | See operation mapping |
+
+**Multi-label nodes.** A Neo4j node can carry several labels at once, for example `(:Person:Customer)`, but a relational row belongs to exactly one table.
```
mfvitale, line 97: Could we just map to `person_customer` as table name by default and eventually give the option to override? 

Also, in the examples I saw around (I'm not an expert) seems that the labels goes from the less general to the more specific. If this is true, we could have another strategy to chose the less or more specific part. 

For example with `Person:Customer` with the generic you will get `person` and with the specific you will get `customer`.

DDD-74.md
```diff
@@ -0,0 +1,542 @@
...
+    "email": "john@foo.org"
+  },
+  "source": {
+    "connector": "neo4j",
+    "ts_ms": 1677844710526,
+    "txId": 12,
+    "table": "Customer"
+  },
+  "op": "c",
+  "ts_ms": 1677844710526
```
mfvitale: Since this normally is the time when the event is processed by Debezium, should we put a different value to highlight it? Or since this is just an STM format conversion we want to keep them aligned?

DDD-74.md
```diff
@@ -0,0 +1,542 @@
...
+    "state": { "before": null, "after": { "properties": { "quantity": 3 } } }
+  },
+  "metadata": { "txCommitTime": "2023-03-03T12:00:00.000Z", "txId": 15 }
+}
+```
+
+Configuration (`column.naming=snake_case` gives the lowercase `order_id` / `product_id` columns; `quantity` comes straight from the property; only the join-table name is overridden, since the default would be the relationship type `CONTAINS`):
+```properties
+transforms.neo4j.column.naming=snake_case
+transforms.neo4j.relationship.CONTAINS.table=order_items
```
mfvitale, line 204: How can I apply multiple mappings for different relations nodes?

DDD-74.md
vsantonastaso: mmm I don't think so. A property-hash key isn't stable since it changes whenever any property changes, so an update would land on a different primary key than the original insert and we could have the same orphaned-row problem as the key value change case. For this reason we won't support the key value change in the V1 as documented in the limitation section

DDD-74.md
vsantonastaso: I think that could be risky since that Neo4j doesn't guarantee label order, so specificity isn't reliable.
My proposal requires that the user has to choose the owning label that is explicit and safe. wDYT?

DDD-74.md
vsantonastaso: As a far as I know in the event we have the `source. ts_ms` that is the commit time and `ts_ms` that is the Debezium processed time. So here we should put a different valua, your're right, thanks

DDD-74.md
vsantonastaso: mmm I add an example to show it for clarification. Thanks

vsantonastaso: Added a dedicated sub section `Reused relationship types (same type, different endpoints)` to explain better this part

vsantonastaso:
Thanks for the review @mfvitale
I replied/addressed you comments 👍

DDD-74.md
mfvitale: If order is not guaranteed then I'm ok.

vsantonastaso: yep unfortunately it is (https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-labels)

image

mfvitale: Wondering how the neo4j source connector get those labels.

vsantonastaso: in the connector doc (https://neo4j.com/docs/cdc/current/procedures/output-schema/) I cannot see anything about the ordering of the labels array, which is already a clue..
But in the neo4j code I see that the labels() returns a TokenSet ([here](https://github.com/neo4j/neo4j/blob/2026.09/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/NodeCursor.java)) and TokenSet is unordered ([here](https://github.com/neo4j/neo4j/blob/2026.09/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/TokenSet.java))
So I suppose that it is unordered

PR #77, opened 2026-09-18
DDD-67: Credentials Management for Debezium Platform

### LINKED ISSUE #67: Debezium Platform credentials management
Provide DDD for pipeline credentials management (secret store integration, OpenBao as reference implementation).

pxcamus:
Closes #67.

Design for credentials management in Debezium Platform: pipeline configuration carries                                     
references instead of secret values, and each pipeline pod fetches its own credentials from a                              
secret store (OpenBao as the reference implementation) using its Kubernetes identity.


DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+  value: "openbao-token"
+- name: PIPELINE_VAULT_TOKEN_EXPIRATION_SECONDS
+  value: "600"
+```
+
+On the secret-store side, the operator enables the Kubernetes auth method once and writes a
+single policy that every pipeline pod shares:
+
+```shell
+bao auth enable kubernetes
```
mfvitale, line 245: Since we are relaxing the requirement to run the Platform only on Kubernetes. Could you explore this possibility also for the credentials?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+a secret _container_; this design repurposes it as a **catalog of references** — records that point at
+credentials the platform itself can never read.
+
+The record changes shape as follows:
+
+| Field | Today | Proposed |
+|---|---|---|
+| `id`, `name`, `description` | as is | unchanged; `name` is what a reference uses: `${vault::<name>/<key>}` |
+| `plaintext` | boolean: the values are stored unencrypted | removed; the platform stores no values |
+| `items` | `Map<String, String>` of secret values | the key names the reference serves, no values |
```
mfvitale, line 327: Should this just be a `List<String>`?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+    Note over P,A: The projected token is stamped for OpenBao only:<br/>replayed against the Kubernetes API it is rejected.
+```
+
+`OpenBaoSecretStore` authenticates via `auth/kubernetes/login` with the
+projected token, then reads the configured path. Response handling covers both
+dynamic engines and KV v2 (values nested under `data.data` beside `data.metadata`).
+
+When `pipeline.vault.enabled: true`, the conductor:
+
+- creates a per-pipeline ServiceAccount (`<pipeline>-sa`, server-side apply,
```
mfvitale, line 287: I would think to let the operator doing this. Effectively the vault reference resolution is done in the Debezium Server so moving that to the operator enables this feature to be used also on just the Operator?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+The record changes shape as follows:
+
+| Field | Today | Proposed |
+|---|---|---|
+| `id`, `name`, `description` | as is | unchanged; `name` is what a reference uses: `${vault::<name>/<key>}` |
+| `plaintext` | boolean: the values are stored unencrypted | removed; the platform stores no values |
+| `items` | `Map<String, String>` of secret values | the key names the reference serves, no values |
+| `path` | absent | new column: the location in the secret store, e.g. `db/ecommerce/creds/pipeline` |
+| store address, auth role | absent | not on the entity: platform-wide configuration (`pipeline.vault.address`, `pipeline.vault.authRole`), one store and one role per platform instance |
+
```
mfvitale, line 330: Could you add an example for an entity?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+Related: today the operator grants every pipeline ServiceAccount read access to all
+Secrets and ConfigMaps in the namespace, including a ServiceAccount the platform supplies
+([debezium/dbz#2327](https://github.com/debezium/dbz/issues/2327)). This design turns off
+token automount, so the pod cannot use that access. Implementing dbz#2327 becomes critical, 
+because only then the access will be totally gone.
+
+### Conductor data model: the Vault entity as a reference catalog
+
+The conductor already carries the schema for this design, but the `Vault` entity today models
```
mfvitale, line 317: This was initially meant to be the container of the plain values, moving that encrypted to the conductor and then based on the provider( secretes, encrypted file, vault manager etc. ) store them. 

We are now moving to outsource the storing part directly to external `SecretStore` service, so this effectively is just a link but seems a bit a duplication of what you already have in the `SecretStore`. 

Maybe we can think as a future enhancement to directly query the `SecretStore`? Do you think it is possible given also the stuff around the IAM?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+
+![Connection form with credentials taken from a vault](DDD-67/connection-create.svg)
+
+**Source form, Filters section.** Picking a connection that uses a vault binds the source to
+that vault; nothing else is asked. The table picker cannot work for such a connection, for the
+same reason `Validate` cannot: it needs a database login, and only the pipeline pod can get one.
+Instead of today's "Failed to load database table/collection" error, the section explains why
+and falls back to typed include and exclude lists. The signal collection check is skipped the
+same way. Nothing breaks silently: a table name that does not exist is reported by the
+pipeline pod when it starts.
```
mfvitale, line 469: To let this working as today, I suppose the conductor must have a connection to the secret store (openbao) to resolve the keys. 

Even we implement this in a another phase, I think we should at least design how it should work.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+
+![Source form Filters section when credentials come from a vault](DDD-67/source-filters.svg)
+
+### Operator responsibilities
```
mfvitale, line 473: `Operator` is a bit confusing here.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+  directly today. A second store needs a selection mechanism, for example a
+  `debezium.vault.<name>.type` property resolved through `ServiceLoader`.
+- **Design-time database access.** Three conductor features open a database connection with
+  the stored credentials: connection validation (`POST /connections/validate`), the table
+  picker in the source Filters section (`GET /connections/{id}/collections`), and the signal
+  collection check. None can work when the credentials are references, because the conductor
+  holds no access to the secret store. The first increment accepts the loss, as shown in the
+  mockups: typed table lists, no validation, errors at pod start. Two candidates to restore
+  them:
+  - Give the conductor its own identity and a separate, metadata-only database role from the
```
mfvitale, line 571: I'll go for this

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+
+## Future Work
+
+- Lease renewal in `OpenBaoSecretStore`: renew the auth token and the secret lease while the
+  pipeline runs, modelled on Spring Vault's `SecretLeaseContainer`, and revoke the held lease
+  in `close()`.
+- Per-pipeline segregation: per-database policies and auth roles, which
+  requires replacing `bound_service_account_names="*"` (per-database
+  namespaces are the variant that avoids a per-pipeline secret-store step).
+- Converting offset storage and schema history JDBC credentials to
```
mfvitale, line 585: Aren't they already cover by the SmallRye config?

mfvitale (commented):
@pxcamus Great work! 

I left some comments.

DDD-67.md
pxcamus: @mfvitale My thinking would be to start with a static token -- env var or a file with `0600` access. It would be a small code change in the OpenBao store. The OpenBao admin can create it easily:

```shell
bao token create -policy=pipeline -ttl=720h -orphan -display-name=dbz-server-vm1
```
Note: better create a token role inside OpenBao and invoke it instead of risking users forget params, etc...

The default max TTL is 768h or 32 days. Once this is in place, the code can incrementally improved to tighten security. WDYT?

DDD-67.md
pxcamus: You are correct, but it will require a database migration. I was keeping the existing shape for now -- explained line 333. I will integrate that to the PR directly then.

DDD-67.md
pxcamus: Agreed, this is the right target, and would indeed enable the feature for people using the operator without the platform.

It would mean a CRD change though for the operator, to add these capabilities:
- Turn off token automount on the SA.
- Mount a projected token with the correct audience.
- The vault settings => `DEBEZIUM_VAULT_*`

I was thinking of doing it in two phases:
1. Keep the wiring in the conductor, to validate e2e the design with the current operator.
2. Add the proper fields to the DebeziumServer CR -- maybe tackle [dbz#2327](https://github.com/debezium/dbz/issues/2327) at the same time, since it touches the same code?

DDD-67.md
pxcamus: Depending on whether I add the database migration right away, here are two examples:

With the existing schema:
```java
{
  "name": "ecommerce",
  "description": "PostgreSQL ecommerce, dynamic role per pod",
  "plaintext": false,
  "items": { "path": "db/ecommerce/creds/pipeline", "keys": "username,password" }
}
```

Proposed:
```java
{
  "id": 1,
  "name": "ecommerce",
  "description": "PostgreSQL ecommerce, dynamic role per pod",
  "path": "db/ecommerce/creds/pipeline",
  "items": ["username", "password"]
}
```

DDD-67.md
pxcamus: I agree, and it would be a good future enhancement. A role granted `list` would see the references but not the secrets. And regarding IAM, with the upcoming RBAC, it becomes even cleaner.

For now, the conductor could log in with its own workload identity, so every DMP user sees the same list. In the future, the user's identity could be used (conductor uses the user's JWT token) so each user only sees what they are allowed to.

A few design questions remain opened:
- the friendly name like `${vault::ecommerce/username}`. Could be metadata in the vault itself?
- other stores: I would have to research which ones grant list without reading the actual secrets. Imagine a store that does not support that: would the entity remain as a fallback then?

DDD-67.md
pxcamus: Sounds good, I will add that with the other changes would you have replied to the other points.

DDD-67.md
pxcamus: This would require a platform chart change. We would need a way to declare a shared vault, since nothing currently attaches a vault to the offset storage (it's no a destination or source). Something like:
```yaml
pipeline:
  vault:
    shared:
      - name: offsets
        path: ...
```

DDD-67.md
mfvitale: So IIUC an alternative to `kubenertes` auth is the `token`. Are you then proposing to start with it for all deployment (k8s and host based) and then eventually add also the others auth if needed?

DDD-67.md
mfvitale: Since this feature is not used, I will go directly with the final design. WDYT?

DDD-67.md
mfvitale: As said in the comment above, I'll go directly with the final design.

DDD-67.md
mfvitale: Well these are implementation details but on the DDD we should document what will be the end design. So we should document here where we want the SA be created. 

IIUC this is required only if the bao auth is kubernetes, and if we decide to start with the `token` approach first. We can just document this in a future section explaining the design to support `kubernetes` auth.

DDD-67.md
pxcamus: I was answering only in the context of your question: how to authenticate Debezium server without the platform (i.e. host based). Many options exist, but to get started a static token is good enough IMHO. 

For k8s with the platform, the Kubernetes auth method is the way to go.

DDD-67.md
mfvitale: For now we can use the `Vault` as you proposed, and we can document a future evolution to directly list them from the external store.

DDD-67.md
mfvitale:
```yaml
offset:
  reusePlatformDatabase: true
  database:
    name: postgres
    host: postgres
    port: 5432
    username: ${vault::ecommerce/username}
    password: ${vault::ecommerce/password}
    auth:
      existingSecret: ""
```

Could't we just let users to specify the vault reference? Effectively the offset will not be created until a pipeline is created. So seems reasonable that users can put the reference and when the platform is started configure that reference. The other option users can still use the secret.

DDD-67.md
mfvitale: Sounds good. So we should add both mode in the DDD then.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+  }
+}'
+
+# 3. Create the source and the destination, and bind each one to the vault it uses.
+#    The "vaults" entry is the binding: it tells the conductor to hand that vault's
+#    coordinates to the pipeline pod. "connection" and "vaults" use the ids from steps 1 and 2.
+curl -s -X POST https://dmp.example.com/api/sources -H 'content-type: application/json' -d '{
+  "name": "pgsql-src", "type": "io.debezium.connector.postgresql.PostgresConnector", "schema": "string",
+  "connection": { "id": 1 },
+  "vaults": [ { "id": 1, "name": "ecommerce" } ],
```
indraraj, line 443: In the UI mockups, I don't see how we're binding the vaults to the source/destination. It looks like we're still relying on the connection entity.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+  "config": {
+    "producer.bootstrap.servers": "my-cluster-kafka-bootstrap.kafka.svc.cluster.local:9094",
+    "producer.security.protocol": "SASL_PLAINTEXT",
+    "producer.sasl.mechanism": "SCRAM-SHA-512",
+    "producer.sasl.jaas.config": "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"${vault::kafka/username}\" password=\"${vault::kafka/password}\";"
+  }
+}'
+
+# 3. Create the source and the destination, and bind each one to the vault it uses.
+#    The "vaults" entry is the binding: it tells the conductor to hand that vault's
```
indraraj, line 438: Also, is this editable in the source/destination later, or will it be a non-editable field?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+
+![Vaults list page](DDD-67/vault-list.svg)
+
+**Add vault.** The form asks for a name, a path and the key names, and nothing else. No secret
+value can be typed anywhere. The panel on the right shows the references the vault will
+provide, ready to copy.
+
+![Add vault form](DDD-67/vault-create.svg)
+
+**Connection form.** Every username, password or token field gets a `Value` / `Vault` switch.
```
indraraj: How do we decide when to add this switch? Should we add it only when the field name matches one of these three specific _`username`, `password` or `token`_ and maintain them manually? Or could this be handled through the schema descriptors instead?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+`Value` keeps today's behaviour. `Vault` replaces the input with a menu of the keys the
+registered vaults serve, grouped by vault, so a reference can only point at a registered vault
+and nobody has to type the `${vault::…}` syntax. The connection then stores the references,
+as in the REST example above. `Validate` is switched off in that case: the platform cannot read
```
indraraj: Since this is effective on the page level I think this `Value` / `Vault` toggle should be present at the page level instead of on the field itself.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+a vault, so it cannot test the credentials, and a wrong reference shows up when the pipeline
+pod starts.
+
+![Connection form with credentials taken from a vault](DDD-67/connection-create.svg)
```
indraraj, line 490: do you think we should support the vault type for the additional properties also?

Image

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+  IAM's 15-minute tokens).
+- **Conductor-mediated custody** (conductor fetches and injects) —
+  breaks the segregation of privileges described at line 303.
+
+## Open Questions
+
+- **Selecting a `SecretStore` implementation.** The factory instantiates `OpenBaoSecretStore`
+  directly today. A second store needs a selection mechanism, for example a
+  `debezium.vault.<name>.type` property resolved through `ServiceLoader`.
+- **Design-time database access.** Three conductor features open a database connection with
```
indraraj: This will also affect the Pipeline action as we recently added the table explorer support for the incremental snapshot action.

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+The proposed flow would look like this:
+
+**Vaults page.** The `Vaults` entry already exists in the navigation behind a feature flag,
+with an empty state only. It becomes a list of references: the name authors use, the path in
+the secret store, the keys it serves, and which sources or destinations are bound to it. The
+store address, auth role and token audience are shown read-only, because they are platform-wide.
+
+![Vaults list page](DDD-67/vault-list.svg)
+
+**Add vault.** The form asks for a name, a path and the key names, and nothing else. No secret
```
indraraj, line 474: In the Vault edit flow which fields can we edit later?

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+
+From there the pipeline is created in the UI as usual, from `pgsql-src` and `kafka-dest`.
+
+### Stage UI changes (mockups)
+
+The proposed flow would look like this:
+
+**Vaults page.** The `Vaults` entry already exists in the navigation behind a feature flag,
+with an empty state only. It becomes a list of references: the name authors use, the path in
+the secret store, the keys it serves, and which sources or destinations are bound to it. The
```
indraraj, line 469: Is delete the action allowed while being in use

DDD-67.md
```diff
@@ -0,0 +1,602 @@
...
+`pipeline.vault.enabled: false` and the conductor emits today's plaintext
+config when it is off. No offset formats, topic naming, or public APIs change.
+
+### Implementation steps
+
+1. `debezium-server`: `SecretStore` SPI, `OpenBaoSecretStore`, SmallRye
+   handler, config model, unit tests. Inert by default.
+2. `debezium-platform`: per-pipeline ServiceAccount, projected token volume, and
+   the bound vaults' coordinates as pod environment, behind
+   `pipeline.vault.enabled`. Tests both ways.
```
indraraj, line 556: Should the Conductor expose something like a feature flag to communicate this to the UI, so the Platform Stage UI can determine whether to enable the Vault feature?

DDD-67.md
pxcamus: @mfvitale can you clarify what you mean please? Starting with a static OpenBao token for the DMP running on Kubernetes in not the right default IMHO. It does not simplify the design, and re-introduces a long-lived static token into the deployment -- it would also make rotation of that token harder. More importantly, the current design relies on the premise of a non-forgeable kubernetes workload identity.

I was thinking that static tokens are only a fallback for a standalone Debezium server. WDYT?

DDD-67.md
pxcamus: @indraraj the binding is shown on the source form, under the Connection field (lock icon, vault chip, "Creating the source binds it to that vault"). It is derived from the connection on purpose: the connection holds the ${vault::…} references, so there is nothing separate for the pipeline author to pick, and a mismatch (source bound to a vault its connection doesn't reference) can't occur. Am I missing something?

DDD-67.md
pxcamus: It will be editable: a source/destination can select a different connection, and it will be re-bound to whatever vault is referenced by that connection. It will follow the normal re-binding process of a pod restart, since references resolve at startup with SmallRye.

DDD-67.md
pxcamus: I definitely recommend going with the descriptors. Different connectors will have different fields to handle, like Kafka: `producer.sasl.jaas.config`, etc...  We could maybe use the field names match as a fallback until each connector's descriptor marks its fields?

@mfvitale WDYT?

DDD-67.md
pxcamus: Good point! I don't think it makes sense to have a hybrid approach, i.e.: `username` is entered manually, but the `password` comes from the vault. WDYT?

DDD-67.md
pxcamus: Agreed, since since resolution is at the config layer, and any property can carry a reference -- JAAS line, transform API keys, etc...

DDD-67.md
pxcamus: Correct, for the first increment of the design: table explorer disabled with an explanation and typed table list fallback. As per @mfvitale feedback, this will be restored later by the conductor's own metadata-only role -- I added that work in the "Future Work" section.

DDD-67.md
pxcamus: `description`, `path` and `items` (keys) should be editable. For `name` we should follow the existing pattern: immutable while bound, since references embed it, mutable otherwise. Same principle for removing a key that a bound connection references: It should be denied.

DDD-67.md
pxcamus: It should not be allowed IMHO.

DDD-67.md
pxcamus: Agreed! We should expose `pipeline.vault.enabled` and other metadata read-only (i.e.: store address, auth role, audience) through a conductor endpoint, so the Stage shows the Vaults page and the Value/Vault switch only when enabled.

DDD-67.md
pxcamus: Agreed on this approach, but the vault named in the reference must still exist in the catalog and the conductor derives the coordinates from it -- same rule as for any source.

DDD-67.md
mfvitale: Yes, is partially what I was saying. We should document in the design both approach, and where they are used. 

> Add the proper fields to the DebeziumServer CR -- maybe tackle https://github.com/debezium/dbz/issues/2327 at the same time, since it touches the same code?

The design of this should be documented here.

DDD-67.md
mfvitale: Could we simplify a bit the things and just show the vault references when users start typing a special key (i.e `/` or `${` )?

In this way we give users the max flexibility.

DDD-67.md
indraraj: Yes, I’m also not inclined toward a hybrid approach. Since if we are already marking a field as sensitive and Vault is being used for other sensitive fields, why would we use manual entry for some sensitive fields

DDD-67.md
mfvitale: See https://github.com/debezium/debezium-design-documents/pull/77#discussion_r4080928986

DDD-67.md
mfvitale: See https://github.com/debezium/debezium-design-documents/pull/77#discussion_r4080928986

DDD-67.md
indraraj: >It is derived from the connection on purpose: the connection holds the ${vault::…} references, so there is nothing separate for the pipeline author to pick, and a mismatch (source bound to a vault its connection doesn't reference) can't occur. 

Yes @pxcamus , that's what I meant to ask that will Vault be derived from the connection itself. thanks 👍

indraraj: > Could we simplify a bit the things and just show the vault references when users start typing a special key (i.e `/` or `${` )?
> 
> In this way we give users the max flexibility.

@pxcamus Correct me if I got it wrong; we won't have a Vault selector directly here. Instead, the Vault value will be shown/derived based on the selected connection. And we already have typeahead support for selecting the connection.

indraraj: > Could we simplify a bit the things and just show the vault references when users start typing a special key (i.e `/` or `${` )?
> 
> In this way we give users the max flexibility.

Yes, we can do that, and I think it would be helpful for the user. Since the Vault will already be locked in(via connection), the user will only need to select the correct key from the available list.

mfvitale: I mean, this should be the general way to use vault for a field. Then in future we can think to add some more advanced UI/UX?

